PR #47, opened 2018-03-29
@carlvitzthum does there already exist a way to store secrets securely in higlass

j1z0:


j1z0:
ok secrets gotten same way as ff secrets now

PR #53, opened 2018-05-11
Chkbiorxiv4pub

aschroed:


chalicelib/wrangler_checks.py
```diff
@@ -11,9 +11,90 @@
...
+                chkdesc = chkdesc + "problem with results format for some records\n"
+            continue
+        ids = result.get('idlist')
+        if ids:
+            # we have possible article(s) - populate check_result
+            fndcnt += 1
+            fulloutput[buuid] = ['PMID:' + id for id in ids]
+
+    if not chkstatus:
+        chkstatus = 'PASS'
```
carlvitzthum: The following four lines will always overwrite `chkdesc`. Intended?

chalicelib/wrangler_checks.py
```diff
@@ -11,9 +11,90 @@
...
+import time
 import boto3
 
 
+@check_function()
+def biorxiv_is_now_published(connection, **kwargs):
+    check = init_check_res(connection, 'biorxiv_is_now_published')
+    chkstatus = ''
+    chkdesc = ''
+    # run the check
```
carlvitzthum: I don't think it's an immediate issue since there aren't many publications, but the `limit=all` search could be dangerous at some point. Maybe add `frame=object` to the url to keep the request size down? Not high priority because I will go through foursight at some point and replace all `limit=all` cases with a safe search.

chalicelib/wrangler_checks.py
aschroed: Nope.  Good catch!

chalicelib/wrangler_checks.py
aschroed: Followed what was used in the hi_glass url check (which BTW seems much more dangerous given the number of processed files but would be happy to switch to safe_search - we maybe should have a shared safe search function then could add specific callback?

PR #68, opened 2018-07-18
Audits

sbreiff:
- added 4 checks that are meant to replace fourfront audits
- fixed url for foursight dev in getting_started.md


chalicelib/checks/wrangler_checks.py
```diff
@@ -472,3 +472,172 @@ def patch_file_size(connection, **kwargs):
...
+def biosource_cell_line_value(connection, **kwargs):
+    '''
+    checks cell line biosources to make sure they have an associated ontology term
+    '''
+    check = init_check_res(connection, 'biosource_cell_line_value')
+
+    cell_line_types = ["primary cell", "primary cell line", "immortalized cell line",
+                       "in vitro differentiated cells", "induced pluripotent stem cell line",
+                       "stem cell", "stem cell derived cell line"]
+    biosources = ff_utils.search_metadata('search/?type=Biosource', ff_env=connection.ff_env, page_limit=200)
```
carlvitzthum: Could this be `search/?type=Biosource&frame=object`? Would be a bit more performant and it doesn't look like the check uses any embedded fields

chalicelib/checks/wrangler_checks.py
```diff
@@ -472,3 +472,172 @@ def patch_file_size(connection, **kwargs):
...
+
+
+@check_function()
+def external_expsets_without_pub(connection, **kwargs):
+    '''
+    checks external experiment sets to see if they are attributed to a publication
+    '''
+    check = init_check_res(connection, 'external_expsets_without_pub')
+
+    ext = ff_utils.search_metadata('search/?award.project=External&type=ExperimentSet',
```
carlvitzthum: Same `frame=object` comment as above

carlvitzthum: As an aside, `page_limit=200` makes some pretty big search requests that have to go through Fourfront. I would recommend setting that value lower, especially if you're getting embedded experiment sets

chalicelib/checks/wrangler_checks.py
```diff
@@ -472,3 +472,172 @@ def patch_file_size(connection, **kwargs):
...
+
+@check_function()
+def expset_opfsets_unique_titles(connection, **kwargs):
+    '''
+    checks experiment sets with other_processed_files to see if each collection
+    of other_processed_files has a unique title within that experiment set
+    '''
+    check = init_check_res(connection, 'expset_opfsets_unique_titles')
+
+    opf_expsets = ff_utils.search_metadata('search/?type=ExperimentSet&other_processed_files.files.uuid%21=No+value',
```
carlvitzthum: For this and the next check, `page_limit` on the search could be set to 20 or something else low

PR #96, opened 2019-01-03
No filesize

aschroed:


chalicelib/checks/wrangler_checks.py
```diff
@@ -468,15 +468,16 @@ def change_in_item_counts(connection, **kwargs):
...
 def identify_files_without_filesize(connection, **kwargs):
     check = init_check_res(connection, 'identify_files_without_filesize')
     # must set this to be the function name of the action
     check.action = "patch_file_size"
-    search_query = ('search/?type=File&status=released%20to%20project'
-                    '&status=released&status=uploaded&frame=object')
-    if kwargs.get('search_add_on'):
-        search_query = ''.join([search_query, kwargs['search_add_on']])
+    default_filetype = 'File'
+    default_stati = 'released%20to%20project&status=released&status=uploaded'
```
sbreiff: maybe also pre-release?

chalicelib/checks/wrangler_checks.py
```diff
@@ -488,13 +489,21 @@ def identify_files_without_filesize(connection, **kwargs):
                 'upload_key': hit.get('upload_key')
             }
             problem_files.append(hit_dict)
+    check.brief_output = '{} files with no file size'.format(len(problem_files))
     check.full_output = problem_files
     if problem_files:
         check.status = 'WARN'
         check.summary = 'File metadata found without file_size'
-        check.description = "One or more files that are released/released to project/uploaded don't have file_size."
+        status_str = 'released/released to project/uploaded'
```
sbreiff: Would we want the status_str to update based on status option? If not then you might still want to add pre-release here.

sbreiff: Oh nevermind i missed the statement below before. But the second part still stands.

PR #97, opened 2019-01-09
checks and actions to generate view configs

Chadius:
# Objective
We want to create higlass view configs for files and experiment sets so "Billy from Bioinformatics" can explore, manipulate and save his own copies.
These checks assume there is a beddb and chromsizes file for each genome assembly, and both files have the "higlass_reference" tag.

Currently if you visit a file with a higlass_uid, one of the tabs will generate an on-the-fly Higlass display. There are a few limitations:
- It only displays one file type (You can see this when files contain different 1D track file types)
- It only displays the first 2D track (Some processed file groups contain multiple mcool files, but only one is shown)
- You can't edit or copy them (a later project will add a "Explore" button that links to the saved viewconfs.)

## Files without higlass_uids
**files_not_registered_with_higlass()** will check for files and note which do not have higlass_uids, which have an invalid higlass_uid, or are not registered on S3.
**patch_file_higlass_uid()** is the corresponding action to create a higlass_uid for the given files.

## Files with higlass_uids
Any File with a higlass_uid should have a viewconf associated with them.

**check_files_for_higlass_viewconf()** looks for files with higlass_uid that are missing viewconf static content:

```
{
        'location': 'tab:higlass_testing',
        'content': view_conf_uuid,
        'description': 'auto_generated_higlass_view_config'
}
```
We'll restore the location to tab:higlass once we confirm it works as expected.

**patch_files_for_higlass_viewconf()** uses the check results to POST new viewconfs (with a title and description) and then PATCHes the new static content.
For testing purposes, you can use file_accession to specify a single file.

## Experiment Sets (ExpSets) and processed files

- The ExpSet's processed_files may have higlass_uids
- The ExpSet's experiments_in_set's processed_files may have higlass_uids

**check_expsets_processedfiles_for_higlass_viewconf()** will find these ExpSets without static content.
**patch_expsets_processedfiles_for_higlass_viewconf()** POSTs the Higlass viewconf and PATCHes it to the ExpSet's static_content section.

You can use the expset_accession flag on both checks to test against a single ExpSet.

## ExpSets and other_processed_files
other_processed_files contains several filegroups. Each filegroup has a title and a list of processed files. For any given other_processed_files section, the titles of each filegroup are unique. For example, only one group will have the title "Archived Files".

The ExpSet's experiments_in_sets may have their own other_processed_files section. Titles between the ExpSet and the Experiments may match. In this case we consider the files part of the same group.

If you visit the ExpSet's page, you can see the "Supplementary Files" section that lists all of the other_processed_files.

**check_expsets_otherprocessedfiles_for_higlass_viewconf()** looks at an ExpSet and its experiments_in_set, trying to collate their other_processed_files containing higlass_uids.
- Each Higlass viewconf will contain the title and combined files from ExpSet and experiments_in_set if the filegroup's title matches.
- If an experiment_in_set has a title but the ExpSet doesn't have it, the ExpSet will get a new filegroup with no files in it.

**patch_expsets_otherprocessedfiles_for_higlass_viewconf()** POSTs the viewconf. It then PATCHes the ExpSet's other_processed_files section to include the viewconf_uuid.

These checks contain my most complicated logic. Making sure all of the old other_processed_files data is untouched before I PATCH it is my biggest concern.
You can use the expset_accession flag on both checks to test against a single ExpSet.

## Deleting cypress test items
I wanted to use cypress tests to make sure the viewconf UI works as expected, but there is no easy way to automate deleting the test viewconfs once they are created. I created a check and action to delete them.

**find_cypress_test_items_to_purge()** This checks for all items that have the tag "deleted_by_cypress_test" and marks them for deletion.
I added a rule priority system just in case this gets more complicated down the road (for example maybe we want to purge files only when they've been deleted for 30 days.)

**purge_cypress_items()** This actually tries to purge tagged items and presents reasons why the purging failed (usually another Item depends on it.)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-dcicutils>=0.5.4
+dcicutils>=0.5.9
```
alexkb0009: 0.6.0 now I think :-P

chalicelib/checks/higlass_checks.py
```diff
@@ -0,0 +1,799 @@
...
+
+    Args:
+        hg_file(bool): A dict representing the file with static content.
+
+    Returns:
+        True if the file has an auto generated higlass view config, False otherwise
+    """
+    # - registered files will have a static_content item with description 'auto_generated_higlass_view_config'.
+
+    # it might be better to check the static_content.location instead...
```
alexkb0009: There could also eventually be some consideration for 'if it exists but is out of date' or similar somehow but not sure. Probably ok to come back to that Q later. Maybe could check date_modified or date_created of all processed_files and if any are more recent than when the HiGlassViewConfig was auto-generated, then it means that expset's viewconfig should be re-generated. Idk just thinking out...keyboard.

chalicelib/checks/higlass_checks.py
```diff
@@ -478,7 +478,7 @@ def patch_expsets_processedfiles_for_higlass_viewconf(connection, **kwargs):
             # Patch the ExpSet static content
             successful_patch, patch_error =  add_viewconf_static_content_to_file(connection, expset_accession, post_viewconf_results["view_config_uuid"],
             static_content_section,
-            "tab:higlass_processed_file_testing")
+            "tab:higlass_processed_files")
```
carlvitzthum: Maybe this tab is supposed to be `supplementary-files`?

e.g. [this exp set](https://data.4dnucleome.org/experiment-set-replicates/4DNESFNBTRO3/#supplementary-files)

carlvitzthum: Maybe this should be `tab:processed-files`?

e.g. this [experiment set tab](https://data.4dnucleome.org/experiment-set-replicates/4DNESXNK2AP5/#processed-files)

chalicelib/checks/higlass_checks.py
```diff
@@ -0,0 +1,1241 @@
...
+    check = init_check_res(connection, 'check_expsets_processedfiles_for_higlass_viewconf')
+    check.full_output = {}
+    check.action = 'patch_expsets_processedfiles_for_higlass_viewconf'
+
+    if kwargs['expset_accession']:
+        expsets_by_accession = {
+            kwargs['expset_accession'] : ff_utils.get_metadata(kwargs['expset_accession'], key=connection.ff_keys, ff_env=connection.ff_env, add_on="frame=embedded")
+        }
+    else:
+        fields_to_include = "&field=" + "&field=".join(
```
carlvitzthum: The first `&field=` is unnecessary. Doesn't seem to break anything but maybe remove

chalicelib/checks/higlass_checks.py
```diff
@@ -0,0 +1,1241 @@
...
+
+    # If an accession was specified, get it
+    if kwargs['expset_accession']:
+        expsets_by_accession = {
+            kwargs['expset_accession'] : ff_utils.get_metadata(kwargs['expset_accession'], key=connection.ff_keys, ff_env=connection.ff_env, add_on="frame=embedded")
+        }
+    else:
+        # Otherwise search for all relevant Experiment Sets
+        # get the fields you need to include
+        fields_to_include = "&field=" + "&field=".join(
```
carlvitzthum: The first `&field=` is unnecessary. Doesn't seem to break anything but maybe remove

chalicelib/checks/higlass_checks.py
```diff
@@ -0,0 +1,1241 @@
...
+
+        if not filetypes_to_use:
+            continue
+
+        # Build a file query string.
+        if file_cat == "raw":
+            type_filter = '&type=FileReference'
+        else:
+            typenames = ['FileProcessed', 'FileVistrack']
+            type_filter = '&type=' + '&type='.join(typenames)
```
carlvitzthum: Remove the first `&type=`. This one will break the search query

chalicelib/checks/higlass_checks.py
```diff
@@ -0,0 +1,1241 @@
...
+        A check/action object.
+    """
+    check = init_check_res(connection, 'files_not_registered_with_higlass')
+    check.status = "FAIL"
+    check.description = "not able to get data from fourfront"
+    # keep track of mcool, bg, and bw files separately
+    valid_filetypes = {
+        "raw": ['chromsizes', 'beddb'],
+        "proc": ['mcool', 'bg', 'bw', 'bed', 'bigbed'],
+    }
```
carlvitzthum: This is minor, but the same information is essentially contained by `valid_filetypes` and `valid_types_raw`/`valid_types_proc`, causing us to have to change multiple spots if we want to add more filetypes. Could you use one or the other? Seems like the only spot the new `valid_filetypes` is used is in line 904: `for file_cat, filetypes in valid_filetypes.items():`

chalicelib/checks/higlass_checks.py
```diff
@@ -0,0 +1,1241 @@
...
+
+        # Only request the necessary fields
+        search_query += "&field=".join(
+            (
+                "accession",
+                "genome_assembly",
+                "file_format",
+                "higlass_uuid",
+                "uuid",
+                "file_format",
```
carlvitzthum: Minor, but `file_format` is duplicated here

Chadius:
To make the query string more obvious, I removed join() and accumulate the string for each term.
valid_types_raw and valid_types_proc are replaced by valid_filetypes.

PR #105, opened 2019-03-12
Combining viewconf check and action for Files

Chadius:
Combining check and action for Files so Higlass viewconfigs are created and patched.

Higlass viewconfig files should be generated for files as quickly as possible. Manually approving and creating viewconfs would be tiresome, so I added the creation to the check so they both run every hour. I added the "patch_files" flag. If it's set to False, no resources will be created or patched.

### DIFF
--- a/chalicelib/checks/higlass_checks.py
+++ b/chalicelib/checks/higlass_checks.py
@@ -225,10 +225,6 @@ def check_files_for_higlass_viewconf(connection, **kwargs):
     headers = {'Content-Type': 'application/json',
                'Accept': 'application/json'}
 
-    # Checks expire after 280 seconds, so keep track of how long this task has lasted.
-    start_time = time.time()
-    time_expired = False
-
     # these are the files we care about
     # loop by genome_assembly
     for ga in target_files_by_ga:
@@ -244,10 +240,6 @@ def check_files_for_higlass_viewconf(connection, **kwargs):
             if time.time() - start_time > 270:
                 time_expired = True
                 break
-
-            # If a particular accession was chosen, skip the others
-            if kwargs['file_accession'] and file_accession != kwargs['file_accession']:
-                continue
 
             # Post a new Higlass viewconf using the file list
             higlass_title = "{acc} - Higlass Viewconfig".format(acc=file_accession)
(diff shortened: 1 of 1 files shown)


chalicelib/checks/higlass_checks.py
```diff
@@ -197,71 +197,49 @@ def check_files_for_higlass_viewconf(connection, **kwargs):
...
     ff_auth = (connection.ff_keys['key'], connection.ff_keys['secret'])
     headers = {'Content-Type': 'application/json',
                'Accept': 'application/json'}
 
-    # pointer to the reference files (by genome_assembly)
-    ref_files_by_ga = gen_check_result['full_output'].get('reference_files', {})
-
     # Checks expire after 280 seconds, so keep track of how long this task has lasted.
     start_time = time.time()
     time_expired = False
```
carlvitzthum: `start_time` is defined earlier; shouldn't be here (I'm guessing it is just a remnant from the action)

chalicelib/checks/higlass_checks.py
```diff
@@ -271,39 +249,42 @@ def patch_files_for_higlass_viewconf(connection, **kwargs):
             if kwargs['file_accession'] and file_accession != kwargs['file_accession']:
```
carlvitzthum: This is handled above, on line 175 `if kwargs['file_accession']:` ...
Doesn't hurt but may be best to remove code that does duplicate things

chalicelib/checks/higlass_checks.py
```diff
@@ -271,39 +249,42 @@ def patch_files_for_higlass_viewconf(connection, **kwargs):
...
-    target_files_by_ga = gen_check_result['full_output'].get('target_files', {})
     file_count = sum([len(target_files_by_ga[ga]) for ga in target_files_by_ga])
-    action.progress = "Created Higlass viewconfs for {completed} out of {possible} files".format(
+    check.summary = "Created Higlass viewconfs for {completed} out of {possible} files".format(
         completed=len(action_logs["new_view_confs_by_file"].keys()),
         possible=file_count
     )
-    return action
+    check.description = check.summary
+    check.allow_action = False
```
carlvitzthum: Not needed -- `check.allow_action` defaults to False; won't do anything if there isn't a check.action set anyways

Chadius:
Replaced by https://github.com/4dn-dcic/foursight/pull/121

PR #107, opened 2019-03-18
added opf status check to audit_checks

sbreiff:
- checks if any files in a set of other_processed_files have mismatching statuses
- checks if a set of other_processed_files has a higher status than parent experiment_set

### DIFF
--- a/chalicelib/checks/audit_checks.py
+++ b/chalicelib/checks/audit_checks.py
@@ -470,6 +470,14 @@ def check_status_mismatch(connection, **kwargs):
 
 @check_function(id_list=None)
 def check_opf_status_mismatch(connection, **kwargs):
+    '''
+    Check to make sure that collections of other_processed_files don't have
+    status mismatches. Specifically, checks that (1) all files in an
+    other_processed_files collection have the same status; and (2) the status of
+    the experiment set is on the same status level or higher than the status of
+    files in the other_processed_files collection (e.g., if the other_processed_files
+    were released when the experiment set is in review by lab.)
+    '''
     check = init_check_res(connection, 'check_opf_status_mismatch')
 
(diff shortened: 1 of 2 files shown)


chalicelib/checks/audit_checks.py
```diff
@@ -468,6 +468,61 @@ def check_status_mismatch(connection, **kwargs):
...
+        for title in titles:
+            file_list = [item['@id'] for fileset in result.get('other_processed_files', [])
+                         for item in fileset['files'] if fileset['title'] == title]
+            file_list.extend([item['@id'] for exp in result.get('experiments_in_set', [])
+                              for fileset in exp['other_processed_files']
+                              for item in fileset['files'] if fileset['title'] == title])
+            statuses = set([status_dict[f] for f in file_list])
+            if len(statuses) > 1:
+                problem_dict[title] = {f: status_dict[f] for f in file_list}
+            elif 'release' not in result['status'] and (
```
aschroed: I think this is because 'released' and 'released to project' have the same status level, correct?  I have run into similar and I think we might generally simplify things if we set 'released to project' and 'pre-release' status levels as intermediate?  

Here is the status_level dictionary I added to the item release script that may do the trick but have not tested it in this case and other status audits as yet.  Might this be a useful approach? I guess it needs the 'archived' and 'archived to project' status levels tweaked for this purpose. 

STATUS_LEVEL = {
    # standard_status
    "released": 10, "current": 10,
    "released to project": 9,
    "pre-release": 8,
    "planned": 6, "submission in progress": 6,
    "in review by lab": 4,
    "revoked": 0, "archived": 0,"deleted": 0, "obsolete": 0, "replaced": 0, "archived to project": 0,
    # additional file statuses
    'to be uploaded by workflow': 4, 'uploading': 4, 'uploaded': 4, 'upload failed': 4, 'draft': 4, 'released to lab': 4}

PR #108, opened 2019-03-18
Valid gene chk

aschroed:


chalicelib/checks/wrangler_checks.py
```diff
@@ -710,3 +712,54 @@ def patch_wfr_and_log(wfr, full_output):
...
+                problems[gid] = 'ncbi timeout'
+        try:
+            rtxt = resp.text
+        except AttributeError:
+            problems[gid] = 'empty response'
+        else:
+            if rtxt.startswith('Error'):
+                problems[gid] = 'not a valid geneid'
+    if problems:
+        check.brief_output = "{} problematic entrez gene ids.".format(len(problems))
```
sbreiff: This might be better as a check.summary than check.brief_output, but it's up to you.

chalicelib/checks/wrangler_checks.py
```diff
@@ -710,3 +712,54 @@ def patch_wfr_and_log(wfr, full_output):
...
+
+
+@check_function()
+def validate_entrez_geneids(connection, **kwargs):
+    ''' query ncbi to see if geneids are valid
+    '''
+    check = init_check_res(connection, 'validate_entrez_geneids')
+    problems = {}
+    timeouts = 0
+    search_query = 'search/?type=Gene&limit=all'
```
sbreiff: could add `&field=geneid` just to grab less info (it doesn't seem like any other fields are being used). we don't have a lot of targets and probably won't have a lot of genes for the near future, but still probably makes the request quicker. just a suggestion.

PR #109, opened 2019-03-22
more_higlass_checks: Adding other processed file checks.

Chadius:
Looking at foursight, I never added other_processed_files checks to the daily schedule.
Also, fixing a typo on one of the check names.


Chadius:
Removing the duplicate check. Must have been introduced from a merge conflict.

PR #121, opened 2019-04-05
Higlass processed file combine

Chadius:
This task will combine the check and action for generating Higlass displays for files.
- Adding check_only flag so this will not create new Higlass displays.
- Fixing bug where every file had to have contributing labs.
- If the reference files are missing, we won't try to make a Higlass viewconf.


Chadius:
We will auto run actions with a different method. Closing.

PR #125, opened 2019-04-16
Use search queries for Higlass checks

Chadius:
Currently, the Higlass Foursight checks look for Files and ExpSets without Higlass items. This task adds additional checks. Here's a summary of the checks that generate Higlass items.

### Processed Files (not Experiment Sets)
- Has a higlass_uid but no Higlass Item in its static_content (existing behavior)
- Higlass Item was modified more than 1 minute before the file was updated
- A custom list of search queries. Any file that matches the query will be forcibly updated. This check must be manually run.

### The Processed Files in Experiment Sets
- Has at least one Processed File with a higlass_uid but no static content Higlass Item in its Processed File static_content (existing behavior)
- Higlass Item was modified more than 1 minute before the Processed Files, Experiments in the set, or Processed Files in the Experiments were updated
- A custom list of search queries. Any ExpSet that matches the query will be forcibly updated. This check must be manually run.

### Other Processed Files (aka Supplementary FIles) in Experiment Sets
- The existing behavior:
-- There is one opf group in the ExpSet with one file with a higlass_uid, but the opf group has no Higlass Item.
-- There is one opf group in the Experiments in the set with a file with a higlass_uid, but there is no corresponding opf group (or there is, but it has no Higlass Item.)
-- (New with this task) The Higlass Item's last modified date is 1 minute before the ExpSet's last modified date. (This relies on https://github.com/4dn-dcic/fourfront/pull/1076)
- Added a second check to add any custom query. Any ExpSet that matches the query will be forcibly updated. This check must be manually run.

## Checks can compare Relative to action completed timestamp
If you include "<get_latest_action_completed_date>"  in the query, we'll replace it with the time the action completed. The parameter `minutes_leeway` lets you fine tune the timestamp. This is useful when the Higlass Items may take minutes to be modified. The action adds a "completed_timestamp" field to its results.

## Added more scheduling
'hourly_checks_2' and 'daily_checks_2' run 15 minutes after 'hourly_checks' and 'daily_checks', respectively. This allows the checks to run without stepping on each other.


Chadius:
The checks that look for modified timestamps are expected to fail the first time for the reason you stated. After that the action completed and the second run will find the timestamp in its results.

I'll try to use the current time as a fallback. `minutes_leeway` can be used to adjust the time from there. I don't expect many records to be marked as updated in that case. so this check will still need to be manually run once.

PR #128, opened 2019-04-22
Upd pending lab action

aschroed:
Add viewing_groups for user if based on those of the lab PI.


chalicelib/checks/wrangler_checks.py
```diff
@@ -827,6 +828,8 @@ def finalize_user_pending_labs(connection, **kwargs):
     action_logs = {'patch_failure': [], 'patch_success': []}
     for user in check_res.get('full_output', []):
         patch_data = {'lab': user['pending_lab']}
+        if 'lab_PI_viewing_groups' in user:
```
carlvitzthum: Since `user` will have `lab_PI_viewing_groups` value of None if the lab doesn't have a PI, I think this if-statement should probably be `if user.get('lab_PI_viewing_groups'):`

carlvitzthum: ^^ Otherwise, patching with None value for viewing groups will give validation error

aschroed: You are correct sir!  Forgot that fields in pending user got a default value of None.  Fixed.

PR #138, opened 2019-05-02
Biofeat org chk

aschroed:
Audit check to find BioFeatures that:

- lack organism_name - tries to guess the appropriate value

- have mismatched organism_name 
       * if the assigned name is less specific that the guess patch it
       * if the assigned name is more specific than the guess just increments a count
       * if it looks like a true mismatch report it
      

Action will patch automatically only those cases where name was lacking or less specific than guessed name.

### DIFF
--- a/chalicelib/check_setup.json
+++ b/chalicelib/check_setup.json
@@ -201,6 +201,18 @@
             }
         }
     },
+    "check_bio_feature_organism_name": {
+        "title": "BioFeatures with missing or mismatched check_bio_feature_organism_name",
+        "group": "Audit checks",
+        "schedule": {
+            "morning_checks": {
+                "all": {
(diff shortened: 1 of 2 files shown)


chalicelib/check_setup.json
```diff
@@ -201,6 +201,18 @@
             }
         }
     },
+    "check_bio_feature_organism_name": {
+        "title": "BioFeatures with missing or mismatched check_bio_feature_organism_name",
```
sbreiff: I'm guessing this was a mistaken autocomplete and it ought to be more like "Biofeatures with missing or mismatched organism name"?

aschroed: Yep - thanks!

PR #167, opened 2019-09-03
MARGI Quality metric summary and wf enzyme check

Lmercadom:


chalicelib/checks/helpers/qc_utils.py
```diff
@@ -36,6 +36,43 @@ def tooltip(numVal):
     return res
 
 
+def calculate_qc_metric_margi_pairsqc(file_uuid, key):
+    '''Patching a pairs file object from margi with quality_metric summary'''
+    import pdb; pdb.set_trace()
```
KorayKirli: import needs to be removed

PR #171, opened 2019-09-05
Add ES disk space check

willronchetti:
- Adds a new check on the daily schedule that checks the remaining disk space on all nodes in the ES cluster

### DIFF
--- a/chalicelib/check_setup.json
+++ b/chalicelib/check_setup.json
@@ -1227,5 +1227,16 @@
                 }
             }
         }
+    },
+    "elastic_search_space" : {
+        "title": "ES Disk Space Check",
+        "group": "Elasticsearch checks",
+        "schedule": {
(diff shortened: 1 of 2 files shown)


chalicelib/check_setup.json
```diff
@@ -1227,5 +1227,16 @@
...
             }
         }
+    },
+    "elastic_search_space" : {
+        "title": "ES Disk Space Check",
+        "group": "Elasticsearch checks",
+        "schedule": {
+            "morning_checks": {
+                "all": {
+                    "kwargs": {}
```
carlvitzthum: This should be `"kwargs" : {"primary": true}`. The Foursight UI only displays the "primary" result, which is done so that you can test checks without overwriting

willronchetti: Added!

PR #172, opened 2019-09-09
yellow_flag_biosamples check fix

sbreiff:
yellow_flag_biosamples check now properly flags biosamples missing cell_culture_details

### DIFF
--- a/chalicelib/checks/badge_checks.py
+++ b/chalicelib/checks/badge_checks.py
@@ -177,29 +177,32 @@ def yellow_flag_biosamples(connection, **kwargs):
         diff_auth = False
         ploidy = False
         bccs = result.get('cell_culture_details', [])
-        for bcc in bccs:
-            for item in [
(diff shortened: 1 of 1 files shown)


chalicelib/checks/badge_checks.py
```diff
@@ -177,29 +177,32 @@ def yellow_flag_biosamples(connection, **kwargs):
...
-            ]:
-                if not bcc.get(item):
-                    messages.append('Biosample missing {}'.format(item))
-            if bcc.get('karyotype'):
-                karyotype = True
-            for protocol in bcc.get('authentication_protocols', []):
-                protocol_item = ff_utils.get_metadata(protocol['@id'], key=connection.ff_keys)
-                auth_type = protocol_item.get('protocol_classification')
-                if not karyotype and auth_type == 'Karyotype Authentication':
+        if not bccs:
```
aschroed: Does this mean that things like tissues and maybe primary cells that aren't actually cultured will get flagged?  

I'm not sure what goes on in this function 'compare_badges_and_messages' so maybe that is where non relevant Biosources get filtered?

PR #173, opened 2019-09-09
Login redirection fix

willronchetti:
- Logging in on Foursight will now redirect you back to the page you were on instead of the default page.

### DIFF
--- a/app.py
+++ b/app.py
@@ -82,7 +82,7 @@ def auth0_callback():
     domain, context = get_domain_and_context(req_dict)
     # extract redir cookie
     cookies = req_dict.get('headers', {}).get('cookie')
-    redir_url = None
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -80,7 +80,14 @@ def auth0_callback():
     request = app.current_request
     req_dict = request.to_dict()
     domain, context = get_domain_and_context(req_dict)
-    resp_headers = {'Location': context + 'view/' + DEFAULT_ENV}
+    # extract redir cookie
+    cookies = req_dict.get('headers', {}).get('cookie')
+    redir_url = None
```
carlvitzthum: Maybe set `redir_url = context + 'view/' + DEFAULT_ENV` so that the homepage redirection works as a fallback?

willronchetti: Good point. Done.

PR #179, opened 2019-09-23
Elasticsearch v1

willronchetti:
1. Implements an Elasticsearch Connection object
2. Refactors various areas to utilize the ES connection to speed up load times on main and history pages.
3. Documentation updates 
4. Phase out `init_*_res` + other cleanups/refactors
5. Some skipped tests still need addressing


chalicelib/run_result.py
```diff
@@ -1,34 +1,111 @@
...
+        """
+        if not self.es:
+            return None
+        return self.connections['es'].get_object(key)
+
+    def get_object(self, key):
+        """
+        Gets an object given a key from the data store.
+        """
+        obj = self.get_es_object(key)
```
willronchetti: use fsconnection to get/put since it is the same

chalicelib/run_result.py
```diff
@@ -247,58 +302,58 @@ def filename_to_datetime(self, key):
...
-        prefixlen = len(self.name) + 1
-        keydatestr = key[prefixlen:-5] # Remove prefix and .json from key.
-        return datetime.datetime.strptime(keydatestr, '%Y-%m-%dT%H:%M:%S.%f')
-
-
+        try:
+            prefixlen = len(self.name) + 1
+            keydatestr = key[prefixlen:-5] # Remove prefix and .json from key.
+            return datetime.datetime.strptime(keydatestr, '%Y-%m-%dT%H:%M:%S.%f')
+        except:
```
willronchetti: this is no longer necessary

chalicelib/run_result.py
```diff
@@ -323,6 +378,30 @@ def format_result(self, uuid):
...
+        is reported
+        """
+        error_msg = None
+        store_method = getattr(self, 'store_result', None)
+        if not callable(store_method):
+            error_msg = 'Do not overwrite the store_result method.'
+        def _validate(attr, t):
+            if not isinstance(attr, t):
+                nonlocal error_msg
+                error_msg = '%s is not of type %s' % (str(attr), str(t))
```
willronchetti: append all messages

chalicelib/run_result.py
```diff
@@ -370,6 +449,28 @@ def format_result(self, uuid):
...
+        _validate(self.status, str)
+        _validate(self.description, str)
+        _validate(self.name, str)
+        _validate(self.output, dict)
```
willronchetti: dont validate this, its unmapped anyway

chalicelib/run_result.py
```diff
@@ -351,13 +430,13 @@ class ActionResult(RunResult):
...
-    def __init__(self, s3_connection, name):
-        self.description = None
+    def __init__(self, connections, name):
+        self.description = ''
         # valid values are: 'DONE', 'FAIL', 'PEND'
         # start with PEND as the default status
         self.status = 'PEND'
-        self.output = None
-        super().__init__(s3_connection, name)
+        self.output = {}
```
willronchetti: Change back to None

migration.py
```diff
@@ -0,0 +1,92 @@
+import app
```
willronchetti: add to scripts directory

migration.py
```diff
@@ -0,0 +1,92 @@
...
+        except:
+            return False
+
+    # if primary is set to true in file then we keep them
+    if PRIMARY:
+        keys_to_delete = list(filter(is_not_primary, keys_to_delete))
+
+    print("Total keys: %s" % len(keys))
+    print("Keys to delete: %s" % len(keys_to_delete))
+    for i in range(0, len(keys_to_delete), 1000):
```
willronchetti: this should use the check like below

tests/test_action_result.py
```diff
@@ -6,10 +6,10 @@ class TestActionResult():
     connection = app_utils.init_connection(environ)
 
     def test_action_result_methods(self):
-        action = run_result.ActionResult(self.connection.s3_connection, self.act_name)
+        action = run_result.ActionResult(self.connection, self.act_name)
         res = action.store_result()
         assert (res.get('status') == 'PEND')
-        assert (res.get('output') is None)
+        assert (res.get('output') == {})
```
willronchetti: This will now need to change

tests/test_check_runner.py
```diff
@@ -4,7 +4,7 @@ def delay_rerun(*args):
     time.sleep(90)
     return True
 
-pytestmark = [pytest.mark.flaky(rerun_filter=delay_rerun)]
+#pytestmark = [pytest.mark.flaky(rerun_filter=delay_rerun)]
```
willronchetti: uncomment

migration.py
willronchetti: this is irritating to do because it would requite me to iterate over all check names and initiate the check for each one to be cleaned, going to stick with this for now

chalicelib/checks/es_checks.py
```diff
@@ -0,0 +1,100 @@
...
+
+@check_function()
+def clean_s3_es_checks(connection, **kwargs):
+    """
+    Cleans old checks from both s3 and es older than one month. Must be called
+    from a specific check as it will take too long otherwise.
+    """
+    check_to_clean = kwargs.get('to_clean')
+    time_limit = 270 if kwargs.get('timeout') is None else kwargs.get('timeout')
+    if not check_to_clean:
```
carlvitzthum: `action` is not defined yet

chalicelib/checks/es_checks.py
```diff
@@ -0,0 +1,100 @@
...
+        action.status = 'FAIL'
+        action.output = action_logs
+        return action
+    action = ActionResult(connection, check_to_clean)
```
carlvitzthum: This should be `CheckResult` now that this is a check. Also may want to change variable name to `check`

chalicelib/checks/es_checks.py
```diff
@@ -0,0 +1,100 @@
...
+    action_logs = {'time out': False}
+    s3 = connection.connections['s3']
+    es = connection.connections['es']
+    one_month_ago = datetime.datetime.utcnow() - datetime.timedelta(days=30)
+    # XXX: how to enforce timeout?
```
carlvitzthum: Off the top of my head, I can think of a few ways to manage the timeout:
1. Use a lower level RunResult method that returns a generator to delete results (which itself would be called in `delete_results())
2. Create a new "timeout" kwarg for `RunResult.delete_results` and implement there
3. Use a signal.

Feel free to come up with an alternate solution, but I think 1. would be best. Would involve defining a new RunResult method, that could optionally take "chunk size" as a kwarg. Let me know if you have questions

chalicelib/checks/es_checks.py
```diff
@@ -0,0 +1,100 @@
...
+    try:
+        n_deleted_s3, n_deleted_es = action.delete_results(prior_date=one_month_ago)
+    except:
+        action_logs['timeout'] = True
```
carlvitzthum: Same feedback as above... A timeout isn't really a FAIL. Also, it is `action_logs['timeout']` here but you use the key `time out` earlier (with a space)

chalicelib/es_connection.py
```diff
@@ -0,0 +1,279 @@
...
+    indices for each one, such as 'foursight-dev-cgap', 'foursight-dev-data' etc
+
+    ESConnection is intended to work with only a single index.
+
+    Implements the AbstractConnection 'interface'
+    """
+    def __init__(self, index=None, doc_type='result'):
+        self.es = es_utils.create_es_client(HOST, use_aws_url=True)
+        self.index = index
+        self.mapping = {} # only needed if we create the index
```
carlvitzthum: If `self.mapping` is a property that is only used in `self.create_index`, can we just make it local to that method?

chalicelib/es_connection.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def get_size(self):
+        """
+        Returns the number of items indexed on this es instance. Returns -1 in
+        failure.
+        """
+        try:
+            return self.es.count(self.index).get('count')
+        except:
+            return -1
```
carlvitzthum: Maybe 0 would make more sense here? It is falsey whereas -1 is not (e.g. `-1 != False`)

chalicelib/fs_connection.py
```diff
@@ -42,3 +46,24 @@ def __init__(self, fs_environ, fs_environ_info, test=False):
...
+        """
+        Queries ES for key - checks S3 if it doesn't find it
+        """
+        obj = None
+        if self.connections['es'] is not None:
+            obj = self.connections['es'].get_object(key)
+        if obj is None:
+            obj = self.connections['s3'].get_object(key)
+            if self.connections['es'] is not None:
+                self.connections['es'].put_object(key, obj)
```
carlvitzthum: Feel free to disagree, but I don't think `get_object` should attempt to PUT to ES. An example would be the action record, which seems reasonable to obtain with `FSConnection.get_object`

I get how this provides a backup option to add missing results to ES, but it seems to be a possibly hard-to-debug source of trouble in the future. I think we can say the only things that add items to ES are `RunResult.store_formatted_result` and the migration check

chalicelib/fs_connection.py
```diff
@@ -42,3 +46,24 @@ def __init__(self, fs_environ, fs_environ_info, test=False):
...
+        return obj
+
+    def put_object(self, key, value):
+        """
+        Puts an object onto both ES and S3
+        """
+        if self.connections['es'] is not None:
```
carlvitzthum: Could this be simplified/generalized for all connections? Something like:
```
for conn in self.connections.values():
    conn.put_object(key, value)
```

chalicelib/mapping.json
```diff
@@ -0,0 +1,172 @@
...
+                "ignore_above": 256
+              }
+            }
+          },
+          "full_output": {
+            "enabled": false,
+            "type": "object",
+            "include_in_all": false
+          },
+          "kwargs": {
```
carlvitzthum: Ignoring `kwargs` in the mapping may be preferable. That said, there are certain ubiquitous kwargs that could be useful to search by (such as "primary"). We could experiment with removing any kwargs that aren't standard; that is, create mappings for some of them and then, since dynamic mapping is turned off, newly encountered ones wouldn't be mapped. We'd have to test to see if this works

chalicelib/checks/es_checks.py
```diff
@@ -0,0 +1,100 @@
...
+    s3 = connection.connections['s3']
+    es = connection.connections['es']
+    n_s3_keys = s3.get_size()
+    n_es_keys = es.get_size()
+    difference = n_s3_keys - n_es_keys
+    full_output = {}
+    full_output['n_s3_keys'] = n_s3_keys
+    full_output['n_es_keys'] = n_es_keys
+    full_output['difference'] = difference
+    if difference > 1000:
```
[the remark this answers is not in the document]
willronchetti: Fair enough, I suppose there isn't much point in allowing it if there are <100 checks to be migrated.

chalicelib/fs_connection.py
willronchetti: Definitely a good point, I see how this could potentially cause problems so I've removed this. My thinking was that in this case the ES connection works like a cache - so in the case that you do request something that isn't in ES chances are you'll want to request it again and it should be moved to ES regardless of the migration schedule.

chalicelib/run_result.py
```diff
@@ -247,58 +297,58 @@ def filename_to_datetime(self, key):
...
-        prefixlen = len(self.name) + 1
-        keydatestr = key[prefixlen:-5] # Remove prefix and .json from key.
-        return datetime.datetime.strptime(keydatestr, '%Y-%m-%dT%H:%M:%S.%f')
-
-
+        try:
+            prefixlen = len(self.name) + 1
+            keydatestr = key[prefixlen:-5] # Remove prefix and .json from key.
+            return datetime.datetime.strptime(keydatestr, '%Y-%m-%dT%H:%M:%S.%f')
+        except:
```
carlvitzthum: Didn't you mean to remove this?

chalicelib/run_result.py
```diff
@@ -1,34 +1,99 @@
...
+        return self.get_object(result_key)
+
+    def get_all_results(self):
+        """
+        Return all results for this check. Should use with care
+        """
+        all_results = []
+        s3_prefix = ''.join([self.name, '/'])
+        relevant_checks = self.list_keys(records_only=True, prefix=s3_prefix)
+        if not relevant_checks:
```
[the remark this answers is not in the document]
willronchetti: If I don't fallback to S3 (both here and in `get_closest_result`) then `test_check_result_methods` fails. It gives no results from ES, even if I explicitly refresh the index.

chalicelib/run_result.py
```diff
@@ -370,6 +444,27 @@ def format_result(self, uuid):
...
+        is reported
+        """
+        error_msg = ''
+        store_method = getattr(self, 'store_result', None)
+        if not callable(store_method):
+            error_msg = 'Do not overwrite the store_result method.'
+        def _validate(attr, t):
+            if not isinstance(attr, t):
+                nonlocal error_msg
+                error_msg += '%s is not of type %s' % (str(attr), str(t))
```
carlvitzthum: Missing newline when adding to `error_msg`

chalicelib/s3_connection.py
```diff
@@ -30,10 +33,37 @@ def get_object(self, key):
...
+        """
+        now = datetime.datetime.utcnow()
+        resp = self.cw.get_metric_statistics(Namespace='AWS/S3',
+                                             MetricName='BucketSizeBytes',
+                                             Dimensions=[
+                                            {'Name': 'BucketName', 'Value': self.bucket},
+                                            {'Name': 'StorageType', 'Value': 'StandardStorage'}],
+                                            Statistics=['Average'],
+                                            Period=86400,
+                                            StartTime=(now-datetime.timedelta(days=1)).isoformat(),
```
carlvitzthum: What is the significance of `StartTime` and `EndTime` here? Is it the average size over the past day as reported by Cloudwatch? Or the size of all objects created in the last day? Should probably add to docstring

willronchetti: Explanation has been added to the docstring.

PR #181, opened 2019-09-25
Upgrade Chalice + Fixes

willronchetti:
- Upgrades Chalice to 1.11.1 and makes some compatibility changes
- Bug fixes from Elasticsearch update
- Enabling checks for S3/ES count and check cleaning
- Adds placeholder check on UI for checks that have no results

### DIFF
--- a/chalicelib/app_utils.py
+++ b/chalicelib/app_utils.py
@@ -137,7 +137,7 @@ def check_authorization(request_dict, env=None):
                 user_res = ff_utils.get_metadata('users/' + payload.get('email').lower(),
                                             ff_env=env_info[e]['ff_env'], add_on='frame=object')
                 if not ('admin' in user_res['groups'] and payload.get('email_verified')):
-                    # fully authorized
+                    # if unauthorized for one, unauthorized for all
                     return False
             return True
         except:
--- a/chalicelib/app_utils.py
+++ b/chalicelib/app_utils.py
@@ -131,11 +131,9 @@ def check_authorization(request_dict, env=None):
                 return False  # we have no env to check auth
             # leeway accounts for clock drift between us and auth0
             payload = jwt.decode(token, b64decode(auth0_secret, '-_'), audience=auth0_client, leeway=30)
-            envs = env.split(',') # handle multiple envs
-            for e in envs:
-                env_info = init_environments(e)
+            for env_info in init_environments(env).values():
                 user_res = ff_utils.get_metadata('users/' + payload.get('email').lower(),
-                                            ff_env=env_info[e]['ff_env'], add_on='frame=object')
+                                            ff_env=env_info['ff_env'], add_on='frame=object')
                 if not ('admin' in user_res['groups'] and payload.get('email_verified')):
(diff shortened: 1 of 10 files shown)


chalicelib/es_connection.py
```diff
@@ -207,7 +208,26 @@ def get_main_page_checks(self, primary=True):
...
+        raw_result = self.search(search)
+        if checks is not None:
+            # figure out which checks we didn't find, add a placeholder check so
+            # that check is still rendered on the UI
+            found = {}
+            for check_name in checks:
+                found[check_name] = False
+            for result in raw_result:
+                found[result['name']] = True
+            for check_name, has_record in found.items():
```
carlvitzthum: This can be simplified to something like:
```
found_checks = [res['name'] for res in raw_result]
for check_name in checks:
    if check_name not in found_checks:
        < create and add placedholder >
```

chalicelib/es_connection.py
```diff
@@ -181,7 +182,7 @@ def get_result_history(self, prefix, start, limit):
         search.update_from_dict(doc)
         return self.search(search)
 
-    def get_main_page_checks(self, primary=True):
+    def get_main_page_checks(self, checks=None, primary=True):
```
carlvitzthum: Please add a test to `test_es_connection` that leverages `get_main_page` with checks argument

chalicelib/es_connection.py
willronchetti: Is the n^2 runtime not a big deal here?

chalicelib/app_utils.py
```diff
@@ -131,12 +131,15 @@ def check_authorization(request_dict, env=None):
                 return False  # we have no env to check auth
             # leeway accounts for clock drift between us and auth0
             payload = jwt.decode(token, b64decode(auth0_secret, '-_'), audience=auth0_client, leeway=30)
-            env_info = init_environments(env)
-            user_res = ff_utils.get_metadata('users/' + payload.get('email').lower(),
-                                            ff_env=env_info[env]['ff_env'], add_on='frame=object')
-            if 'admin' in user_res['groups'] and payload.get('email_verified'):
-                # fully authorized
-                return True
+            envs = env.split(',') # handle multiple envs
```
carlvitzthum: "all" is a valid foursight environment, which will not work with the code as written. Easy fix:
```
# will handle "all" as well as comma separated envs
for environment in init_environments(env):
```

chalicelib/es_connection.py
carlvitzthum: I guess you could use a `set` to store `found_checks` in the code I wrote so the lookup is faster. But it's up to you, I'm cool with the old code if you want to keep it

chalicelib/app_utils.py
willronchetti: Fixed and now being tested

chalicelib/es_connection.py
willronchetti: Going with set because its cleaner

PR #184, opened 2019-10-01
main page optimization, make favicon configurable

willronchetti:
- No longer repeat calls to `init_environments` on main page through `init_connection`
- Make Favicon configurable from the backend

### DIFF
--- a/chalicelib/app_utils.py
+++ b/chalicelib/app_utils.py
@@ -88,7 +88,7 @@ def init_connection(environ):
     if environ not in environments:
         error_res = {
             'status': 'error',
-            'description': 'invalid environment provided. Should be one of: %s' % (str(list(environments.keys()))),
+            'description': 'environment %s is not valid!' % environ,
             'environment': environ,
             'checks': {}
         }
--- a/chalicelib/app_utils.py
+++ b/chalicelib/app_utils.py
@@ -8,6 +8,7 @@ import boto3
 import datetime
 import ast
 import copy
+import requests
 from itertools import chain
 from dateutil import tz
 from base64 import b64decode
@@ -38,7 +39,7 @@ from .utils import (
 )
 from .s3_connection import S3Connection
 
-FAVICON_4DN = 'https://data.4dnucleome.org/static/img/favicon-fs.ico'
+DEFAULT_FAVICON = 'https://data.4dnucleome.org/static/img/favicon-fs.ico'
 
 jin_env = Environment(
     loader=FileSystemLoader('chalicelib/templates'),
@@ -75,15 +76,14 @@ def init_environments(env='all'):
     return environments
 
 
-def init_connection(environ, environments=None):
+def init_connection(environ):
     """
     Initialize the fourfront/s3 connection using the FSConnection object
     and the given environment.
     Returns an FSConnection object or raises an error.
     """
     error_res = {}
-    if environments is None:
-        environments = init_environments()
+    environments = init_environments(environ)
     # if still not there, return an error
     if environ not in environments:
         error_res = {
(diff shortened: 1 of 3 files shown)


chalicelib/app_utils.py
```diff
@@ -73,14 +75,15 @@ def init_environments(env='all'):
...
-def init_connection(environ):
+def init_connection(environ, environments=None):
     """
     Initialize the fourfront/s3 connection using the FSConnection object
     and the given environment.
     Returns an FSConnection object or raises an error.
     """
     error_res = {}
-    environments = init_environments()
+    if environments is None:
```
carlvitzthum: Good catch that this is not ideal. However, I don't think this is the way to fix it. What about:
```
def init_connection(environ):
    environments = init_environments(environ)   # will return empty dict if not valid
   ...
```
The rest of the code would be mostly the same, only thing is the error message would have to change because you're not guaranteed to have all environments available. To keep this functionality, you could do:
```
if environ not in environments:
    environments = init_environments('all')  # same as init_environments()
    # continue with error message
    ...
```

chalicelib/app_utils.py
```diff
@@ -38,6 +38,8 @@
 )
 from .s3_connection import S3Connection
 
+FAVICON_4DN = 'https://data.4dnucleome.org/static/img/favicon-fs.ico'
```
carlvitzthum: This is not right. If you are using the "fourfront-cgap" environment, it would continue to get the favicon from Fourfront. The only benefit you get from defining it here (instead of the template) is that you can programmatically change it if you wanted.

Instead, you should pass the current Fourfront/CGAP server (i.e. `connection.ff_server` to the template as a new variable and then render `{{server + '/static/img/favicon-fs.ico'}}` wherever you need the favicon. Or you could add the favicon path to the variable before you pass it to the template; same deal.

If you have multiple environments, that's where it gets tricky. Currently, `view_foursight` is the only templating function that supports multiple environments, in which case I think you can just choose one randomly to get the favicon from. Or use `FAVICON_4DN` as a default. Up to you.

chalicelib/app_utils.py
willronchetti: In the second code block, why call `init_environments` at all there if you're just going to raise an exception immediately after? Should it check again if `all` populated `environ`?

chalicelib/app_utils.py
willronchetti: Is it fair to assume the asset will always be available regardless of what `server` is? How can I best detect if isn't prior to rendering the page? I'd need to know to fallback to `FAVICON_4DN`, which I'll probably rename to `DEFAULT_FAVICON`, before rendering the page, no? I guess I could hardcode the default into the html and if the favicon 404's use that one instead?

My thought for multiple environments would be to just pick the first one and use that.

chalicelib/templates/base.html
```diff
@@ -324,7 +324,7 @@ <h5 class="center-element">
             },
             socialButtonStyle: 'big',
             languageDictionary: { title: "Log in" },
-            theme: { logo: favicon },
+            theme: { logo: "{{favicon}}" },
```
carlvitzthum: I'm not sure that Jinja templating works in <script> tags. You should test logging in to see if the icon appears; if not, you could use a `<meta>` tag to pass in `favicon`, similar to what is being done with `fs_domain`

chalicelib/app_utils.py
```diff
@@ -75,15 +76,14 @@ def init_environments(env='all'):
...
+def init_connection(environ):
     """
     Initialize the fourfront/s3 connection using the FSConnection object
     and the given environment.
     Returns an FSConnection object or raises an error.
     """
     error_res = {}
-    if environments is None:
-        environments = init_environments()
+    environments = init_environments(environ)
```
carlvitzthum: Thanks for fixing, but the "description" in `error_res` still needs to be adjusted. Previously it would should all valid environments, but now it is misleading as written

PR #185, opened 2019-10-01
add time limit to dedup

KorayKirli:


chalicelib/checks/system_checks.py
```diff
@@ -339,7 +342,8 @@ def secondary_queue_deduplication(connection, **kwargs):
 
     exit_reason = 'out of time'
     dedup_msg = 'FS dedup uuid: %s' % kwargs['uuid']
-    while elapsed < time_limit:
+    have_more_time = True
+    while have_more_time:
```
willronchetti: Not a big deal but I would do the below instead.

```suggestion
    while True:
    ...
    if time_limit:
        elapsed = round(time.time() - t0, 2)
        if elapsed < time_limit:
            break
```

PR #192, opened 2019-10-15
Bedtomultivec

Lmercadom:


chalicelib/checks/helpers/wfr_utils.py
```diff
@@ -201,6 +205,20 @@
           'fruit-fly': 'dm6',
           'chicken': 'galGal5'}
 
+# color map states bed file
+states_file_type = {'SPIN_states_v1': {'color_mapper': '/files-reference/4DNFI27WSLAG/', 'num_states': 9}}
+
+
+# check for a specific tag in a states file
+def isthere_states_tag(a_file):
```
willronchetti: `a_file` by name and the comment above implies a file type but looks like its actually a dictionary. I would change the comment and the argument name so it's less ambiguous.

chalicelib/checks/wfr_checks.py
```diff
@@ -1514,3 +1513,123 @@ def margi_start(connection, **kwargs):
...
+    check.action = "bed2multivec_start"
+    check.brief_output = []
+    check.full_output = {}
+    check.status = 'PASS'
+
+    # check indexing queue
+    env = connection.ff_env
+    indexing_queue = ff_utils.stuff_in_queues(env, check_secondary=True)
+    if indexing_queue:
+        check.status = 'PASS'  # maybe use warn?
```
willronchetti: I'd ask Burak maybe on what he would prefer - in my mind `WARN` makes more sense as `PASS` implies the check ran successfully when in fact it really did not.

PR #197, opened 2019-10-22
Cgap pipeline

Lmercadom:


chalicelib/checks/helpers/cgap_utils.py
```diff
@@ -483,3 +487,17 @@ def start_tasks(missing_runs, patch_meta, action, my_auth, my_env, start, move_t
     action.output = action_log
     action.status = 'DONE'
     return action
+
+
+def is_there_my_qc_metric(file_meta, qc_metric_name, my_auth):
+    if not file_meta.get('quality_metric'):
+        return False
+    else:
```
KorayKirli: you don't really need else here, but that is a small detail.

chalicelib/checks/helpers/cgap_utils.py
```diff
@@ -483,3 +487,17 @@ def start_tasks(missing_runs, patch_meta, action, my_auth, my_env, start, move_t
...
+        qc_results = ff_utils.get_metadata(file_meta['quality_metric']['uuid'], key=my_auth)
+        if not qc_results.get('qc_list'):
+            return False
+        else:
```
KorayKirli: again, this else is not needed

PR #210, opened 2019-12-10
Rnaqc

SooLee:
Local tests worked. Checked the results and everything looks fine to me.

### DIFF
--- a/chalicelib/checks/helpers/qc_utils.py
+++ b/chalicelib/checks/helpers/qc_utils.py
@@ -12,7 +12,7 @@ def calculate_qc_metric_rnaseq(file_uuid, key):
     if 'star_log_qc' in quality_metric:
         uniquely_mapped = quality_metric['star_log_qc']['Uniquely mapped reads number']
         multi_mapped = quality_metric['star_log_qc']['Number of reads mapped to multiple loci']
-        total_mapped = uniquely_mapped + multi_mapped
+        total_mapped = int(uniquely_mapped) + int(multi_mapped)
         qc_summary.append({"title": "Total mapped reads (genome)",
                            "value": str(total_mapped),
                            "numberType": "integer"})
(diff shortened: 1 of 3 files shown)


chalicelib/checks/qc_checks.py
```diff
@@ -180,6 +180,64 @@ def identify_files_without_qc_summary_bed(connection, **kwargs):
...
+    check = CheckResult(connection, 'identify_files_without_qc_summary_rnaseq')
+    # must set this to be the function name of the action
+    check.action = 'patch_quality_metric_summary_rnaseq'
+    default_filetype = 'FileProcessed'  # skip fastq
+    default_stati = 'released%20to%20project&status=released&status=uploaded&status=pre-release'
+    filetype = kwargs.get('file_type') or default_filetype
+    stati = 'status=' + (kwargs.get('status') or default_stati)
+    search_query = 'search/?type={}&{}&frame=object'.format(filetype, stati)
+    search_query += '&track_and_facet_info.experiment_type=RNA-seq'
+    search_query += '&file_type=gene+expression&file_type=read+positions'
```
willronchetti: You're free to use `ff_utils.search_metadata` but I would definitely recommend looking at `ff_utils.faceted_search` as it can make building these queries a lot cleaner.

chalicelib/checks/helpers/qc_utils.py
```diff
@@ -1,6 +1,37 @@
 from dcicutils import ff_utils
 
 
+def calculate_qc_metric_rnaseq(file_uuid, key):
```
willronchetti: As an aside, I'm noticing that none of the check related functions in `helpers` are directly tested. It may be worthwhile at some point to test these in some way. This is probably a task for when we split Foursight into two repos.

chalicelib/checks/qc_checks.py
SooLee: Good to know. Thanks! I will merge this one for now since it's working (no more cases to test for now), but for the future I will definitely try faceted_search.

chalicelib/checks/helpers/qc_utils.py
SooLee: Sounds good.

PR #222, opened 2020-01-15
Tweak Check Schedule

willronchetti:
- Changes `purge_download_tracking_items` to no longer bail when staging deploy is running
- Changes the above check to run every 10 mins
- Changes `change_in_item_counts` to run on the secondary morning check schedule


chalicelib/check_setup.json
```diff
@@ -110,7 +110,7 @@
         "title": "Change in item counts",
         "group": "Metadata checks",
         "schedule": {
-            "morning_checks": {
+            "morning_checks_2": {
```
netsettler: Why this rename?  Is this just a name or something semantic?

chalicelib/check_setup.json
```diff
@@ -756,7 +756,7 @@
         "title": "Process Download Tracking Items",
         "group": "System checks",
         "schedule": {
-            "hourly_checks": {
+            "ten_min_checks": {
```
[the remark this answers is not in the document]
willronchetti: Correct. If we want to keep this we need it running very regularly as we can only purge ~100 or so every run. If we do it every hour we will only delete ~2400 and on Cloudwatch I see tens of thousands being generated overnight. This change will allow us to at least purge ~15000 or so a day, which should help us keep up.

chalicelib/check_setup.json
willronchetti: `morning_checks_2` is actually a separate, unused schedule that is staggered from the regular `morning_checks`. You can see the schedule in `app.py`.

PR #223, opened 2020-01-16
Poetry + Lambda Timeout

willronchetti:
- Configure the package to install with poetry
- Up lambda timeout to 15 mins
- Fix autodoc


deploy.py
```diff
@@ -16,7 +16,7 @@
       "api_gateway_stage": "api",
       "autogen_policy": False,
       "lambda_memory_size": 512,
-      "lambda_timeout": 300,
+      "lambda_timeout": 900,
```
netsettler: I would add a comment here explaining the 900...
```
      "lambda_timeout": 900,  # seconds (= 15 minutes)
```

deploy.py
```diff
@@ -25,7 +25,7 @@
...
```
netsettler: Again
```
      "lambda_timeout": 900,  # seconds (= 15 minutes)
```

deploy.py
```diff
@@ -59,6 +59,8 @@ def build_config_and_deploy(stage):
     print(''.join(['Writing: ', filename]))
     with open(filename, 'w') as config_file:
         config_file.write(json.dumps(CONFIG_BASE))
+    # export poetry into requirements
+    subprocess.call(['poetry', 'export', '-f', 'requirements.txt', '>', 'requirements.txt'])
```
netsettler: I just have to say this is bizarre syntax. I looked it up in the doc and it seems right, though. I guess the `requirements.txt` after the `-f` is not a filename but a format name keyword. It seems to me that they should have just called it requirements. Oh well. Anyway, this is fine and what the doc seems to recommend.

pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[tool.poetry]
+name = "foursight"
+version = "0.9.0"
+description = "Serverless Chalice Application for Monitoring"
+authors = ["William Ronchetti <wrr33@cornell.edu>"]
```
netsettler: Small matter, but don't you want your Harvard email here?

deploy.py
willronchetti: It is strange isn't it. I'm worried that this change will completely break the deployment of the chalice application because it relies on `requirements.txt`. This is my workaround that I can't exactly test unless I try to deploy over it. Any thoughts on what to best do are appreciated.

pyproject.toml
willronchetti: Correct. Will fix

.coveragerc
```diff
@@ -1,6 +1,4 @@
-[flake8]
```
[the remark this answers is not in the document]
willronchetti: It's unused but I could hypothetically add it back if we care (or replace it with something else). I find it cumbersome and often just makes me conform to an arbitrary standard (usually line length and newline spacing).

PR #240, opened 2020-02-14
cgap pipeline v11->v12

SooLee:


SooLee:
version upgrade v11->12 for all parts,
cram2fastq is now v13.

PR #247, opened 2020-02-24
C4-46 Multiprocess checks

willronchetti:
- Refactor how checks/actions are run into a 2 process setup so we can survive I/O blocks. Parent runs the check, child runs timeout logic. Whoever 'wins' kills the other.
- The parent process spins off a child process to execute timeout logic. This is so that the parent who is running the check can be killed with `SIGTERM` instead of waking up via `SIGALRM`, which does not appear to work on Lambda servers if the check is executing I/O. 
- This will ensure all Foursight checks correctly report results even in the presence of timeout due to I/O.
- This should be very heavily inspected as it is (relatively) complicated multiprocessing code using the `os` interface. When designing this I found the higher level interfaces too clunky. This task is relatively simple.

### DIFF
--- a/chalicelib/utils.py
+++ b/chalicelib/utils.py
@@ -152,6 +152,7 @@ def check_function(*default_args, **default_kwargs):
                     check.full_output = traceback.format_exc().split('\n')
                 kwargs['runtime_seconds'] = round(time.time() - start_time, 2)
                 check.kwargs = kwargs
+                os.kill(child_pid, signal.SIGKILL)  # we finished, so kill child
                 return check.store_result()
             else:  # we are the child who handles the timeout
                 partials = {'name': func.__name__, 'kwargs': kwargs, 'is_check': True,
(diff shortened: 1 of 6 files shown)


chalicelib/utils.py
```diff
@@ -127,23 +138,27 @@ def check_deco(func):
...
-                check = CheckResult(args[0], func.__name__)
-                check.status = 'ERROR'
-                check.description = 'Check failed to run. See full output.'
-                check.full_output = traceback.format_exc().split('\n')
-            signal.alarm(0)
-            kwargs['runtime_seconds'] = round(time.time() - start_time, 2)
-            check.kwargs = kwargs
-            return check.store_result()
+            parent_pid = os.getpid()
+            child_pid = os.fork()
```
netsettler, line 142: While a subprocess seems reasonable, I'm curious about the choice of `os.fork` vs some sort of `subprocess` choice.

PR #249, opened 2020-02-25
higlass update check should now patch existing viewconf if present

sbreiff:
Currently, there's a check for higlass triggering over and over again because its finding experiment sets that were last updated after the last time the higlass action was run, although this seems to be due to indexing delay. The purpose of this PR is to allow cases where higlass_view_confs are only updated, to happen without patching/modifying the experiment set.


utku-ozturk:
@ugur-camoglu and I checked it, looks and works great! old implementation likely to fail w/ increasing volume of data soon.

PR #257, opened 2020-03-26
C4-96 Case fix

willronchetti:
- Upgrade seems to have made a symmetric change to the application registry, causing that which was snake_case before to become camel_case (and vice versa). 
- Account for this in the `change_in_item_counts` check.

### DIFF
--- a/chalicelib/checks/system_checks.py
+++ b/chalicelib/checks/system_checks.py
@@ -8,7 +8,8 @@ from ..run_result import CheckResult, ActionResult
 from dcicutils import (
     ff_utils,
     es_utils,
-    beanstalk_utils
+    beanstalk_utils,
+    env_utils
 )
 
 import requests
@@ -563,7 +564,7 @@ def snapshot_rds(connection, **kwargs):
         check.summary = check.description = 'This check only runs on Foursight prod'
         return check
     # XXX: must be updated when cgap-blue/cgap-green come to fruition -will 4-1-2020
-    rds_name = 'fourfront-production' if ('green' in connection.ff_env or 'blue' in connection.ff_env) else connection.ff_env
+    rds_name = 'fourfront-production' if (env_utils.is_fourfront_env(connection.ff_env) and env_utils.is_stg_or_prd_env(connection.ff_env)) else connection.ff_env
     # snapshot ID can only have letters, numbers, and hyphens
     snap_time = datetime.datetime.strptime(kwargs['uuid'], "%Y-%m-%dT%H:%M:%S.%f").strftime("%Y-%m-%dT%H-%M-%S")
     snapshot_name = 'foursight-snapshot-%s-%s' % (rds_name, snap_time)
--- a/chalicelib/checks/system_checks.py
+++ b/chalicelib/checks/system_checks.py
@@ -562,7 +562,8 @@ def snapshot_rds(connection, **kwargs):
     if get_stage_info()['stage'] != 'prod':
         check.summary = check.description = 'This check only runs on Foursight prod'
         return check
-    rds_name = 'fourfront-webprod' if 'webprod' in connection.ff_env else connection.ff_env
+    # XXX: must be updated when cgap-blue/cgap-green come to fruition -will 4-1-2020
+    rds_name = 'fourfront-production' if ('green' in connection.ff_env or 'blue' in connection.ff_env) else connection.ff_env
     # snapshot ID can only have letters, numbers, and hyphens
     snap_time = datetime.datetime.strptime(kwargs['uuid'], "%Y-%m-%dT%H:%M:%S.%f").strftime("%Y-%m-%dT%H-%M-%S")
     snapshot_name = 'foursight-snapshot-%s-%s' % (rds_name, snap_time)
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -542,13 +542,14 @@ def change_in_item_counts(connection, **kwargs):
     search_query += '&status=deleted&status=replaced'
     search_resp.extend(ff_utils.search_metadata(search_query, key=connection.ff_keys))
     for res in search_resp:
-        # convert type to index name. e.g. ExperimentSet --> experiment_set
-        res_index = res['@type'][0]
-        if res_index in diff_counts:
-            diff_counts[res_index]['ES'] = diff_counts[res_index]['ES'] + 1 if 'ES' in diff_counts[res_index] else 1
+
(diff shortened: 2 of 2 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -541,15 +541,15 @@ def change_in_item_counts(connection, **kwargs):
...
-            diff_counts[res_index]['ES'] = diff_counts[res_index]['ES'] + 1 if 'ES' in diff_counts[res_index] else 1
+
+        # Stick with given type name in CamelCase since we need it to match ES type-name
+        _type = res['@type'][0]
+        if _type in diff_counts:
+            diff_counts[_type]['ES'] = diff_counts[_type]['ES'] + 1 if 'ES' in diff_counts[_type] else 1
         else:
             # db entry wasn't already present for this index
-            diff_counts[res_index] = {'DB': 0, 'ES': 1}
+            diff_counts[_type] = {'DB': 0, 'ES': 1}
```
netsettler, line 552: Is this the same as:
```
diff_counts[_type]['ES'] = diff_counts[_type].get('ES', 0) + 1
```
Seems like you could almosthave set up `diff_counts` to be a `defaultdict(lambda: {'DB': 0, 'ES': 0}` and then you would not need to have checked. But it depends on whether you want entries with all zeros to show up if nothing gets modified. Since it seems like you do, then perhaps better something like this (untested), which seems to be all that's going on here and is maybe easier to read, assuming I didn't make a mistake:
```
_type = res['@type'][0]
_entry = diff_counts.get(_type)
if not _entry:
    diff_counts[_type] = _entry = {'DB': 0, 'ES': 0}
if _type in diff_counts:
    _entry['ES'] += 1
```

I don't see unit tests in this PR so don't know if this is tested, but given that the shape of things is changing, I'd be fairly surprised if there were unit tests that spanned. I'd expect an existing unit test to be broken and need fixing if there were good ones.

chalicelib/checks/wrangler_checks.py
```diff
@@ -560,7 +560,7 @@ def change_in_item_counts(connection, **kwargs):
     total_counts_db = sum([diff_counts[coll]['DB'] for coll in diff_counts if diff_counts[coll]['DB'] >= 0])
     # see if we have negative counts
     # allow negative counts, but make note of, for the following types
-    purged_types = ['tracking_item', 'higlass_view_config']
+    purged_types = ['TrackingItem', 'HiglassViewConfig']
```
netsettler, line 563: This strongly argues for taking the names of the tags and separating them out into constants. Might be better to do that in an organized way, so perhaps a task for another day. But I just wanted to say it here to have said it. I'd settle for a `# TODO` comment perhaps?

netsettler:
(I wouldn't at all mind seeing better unit tests added, though, whether here or later.)

chalicelib/checks/wrangler_checks.py
netsettler: In case that was opaque, by "if nothing gets modified" I mean "if you never end up bumping the ES count above 0".

chalicelib/checks/wrangler_checks.py
```diff
@@ -531,6 +530,7 @@ def change_in_item_counts(connection, **kwargs):
 
     # now do a metadata search to make sure they match
     # date_created endpoints for the FF search
+    # XXX: We should revisit if we really think this is necessary. - will 3-26-2020
     to_date = datetime.datetime.strptime(latest_check['uuid'], "%Y-%m-%dT%H:%M:%S.%f").strftime('%Y-%m-%d+%H:%M')
     from_date = datetime.datetime.strptime(prior_check['uuid'], "%Y-%m-%dT%H:%M:%S.%f").strftime('%Y-%m-%d+%H:%M')
```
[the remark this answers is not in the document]
willronchetti: Sorry that this wasn't clear - I am not referring to this date change. All that this is doing is figuring out an approximate "window" to search in. What I was referring to was the fact that the search is done at all. It is sufficient to just look at `CheckResult(connection, 'item_counts_by_type')` at the current time and 24 hours prior. We are doing an extra search here that seems unnecessary to me.

chalicelib/checks/wrangler_checks.py
```diff
@@ -541,15 +541,15 @@ def change_in_item_counts(connection, **kwargs):
...
     search_query += '&status=deleted&status=replaced'
     search_resp.extend(ff_utils.search_metadata(search_query, key=connection.ff_keys))
-    search_output = []
     for res in search_resp:
-        # convert type to index name. e.g. ExperimentSet --> experiment_set
-        res_index = convert_camel_to_snake(res['@type'][0])
-        if res_index in diff_counts:
-            diff_counts[res_index]['ES'] = diff_counts[res_index]['ES'] + 1 if 'ES' in diff_counts[res_index] else 1
+
+        # Stick with given type name in CamelCase since we need it to match ES type-name
```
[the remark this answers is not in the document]
willronchetti: This isn't clear from the comment so I'll fix it but the CamelCase-ness comes from the counts page (snovault). We used to have to convert CamelCase to snake_case because snake_case is what we transformed to in snovault. The update caused the symmetric change to now convert snake_case to CamelCase, so the conversion here is now unnecessary.

chalicelib/checks/wrangler_checks.py
willronchetti: There are no unit tests because these are foursight checks, they are only 'tested' when run manually from the terminal or in production (unfortunately). I like your change better though and will take it.

chalicelib/checks/wrangler_checks.py
willronchetti: Agreed, I'll add a comment.

chalicelib/checks/system_checks.py
```diff
@@ -562,7 +562,8 @@ def snapshot_rds(connection, **kwargs):
     if get_stage_info()['stage'] != 'prod':
         check.summary = check.description = 'This check only runs on Foursight prod'
         return check
-    rds_name = 'fourfront-webprod' if 'webprod' in connection.ff_env else connection.ff_env
+    # XXX: must be updated when cgap-blue/cgap-green come to fruition -will 4-1-2020
+    rds_name = 'fourfront-production' if ('green' in connection.ff_env or 'blue' in connection.ff_env) else connection.ff_env
```
netsettler, line 566: Do you not want (untested):
```
rds_name = 'fourfront-production' if is_stg_or_prd_env(connection.ff_env) else connection.ff_env
```
It would need an import from the new `dcicutils.env_utils`.

PR #266, opened 2020-04-16
handling for 404s on ontology update check

sbreiff:
**Problem:**
Today this check errored because the SO download_url was giving a 404. 

**In this PR:**
- 404 handling in ontology check
- assay classification update for optodroplet

### DIFF
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -1324,7 +1324,7 @@ def check_for_ontology_updates(connection, **kwargs):
                 versions[o['ontology_prefix']]['latest'] = 'WARN: 404 at download_url'
                 check.summary = '404 at download_url'
                 check.description = 'One or more ontologies has a download_url with a 404 error.'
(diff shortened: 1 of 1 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -1318,6 +1320,13 @@ def check_for_ontology_updates(connection, **kwargs):
         # instead of repos etc, check download url for ontology header to get version
         elif o.get('download_url'):
             owl = requests.get(o['download_url'], headers={"Range": "bytes=0-2000"})
+            if owl.status_code == 404:
```
aschroed, line 1323: Would it be worth it to have any retries here? Probably not as it's only happened the one time so far and we don't know how quickly it resolved.

sbreiff: In the current case it was a permanent move, not a temp outage. I'll leave as is for now but can add retries if temporary 404s happen in the future.

PR #267, opened 2020-04-21
Added check to report fastqs with integer read ids

Lmercadom:
-I also added restricted status to the fastq_first_line wfr check

### DIFF
--- a/chalicelib/checks/audit_checks.py
+++ b/chalicelib/checks/audit_checks.py
@@ -794,7 +794,7 @@ def check_fastq_read_id(connection, **kwargs):
     check.summary = 'No fastq files with integer ids'
     check.full_output = {}
     check.status = 'PASS'
-    query = '/search/?date_created.from=2019-01-01&file_format.file_format=fastq&status=uploaded&type=FileFastq'
+    query = '/search/?date_created.from=2020-04-13&file_format.file_format=fastq&status=uploaded&type=FileFastq'
     res = ff_utils.search_metadata(query, key=connection.ff_keys)
     if not res:
         return check
--- a/chalicelib/checks/audit_checks.py
+++ b/chalicelib/checks/audit_checks.py
@@ -782,3 +782,32 @@ def patch_bio_feature_organism_name(connection, **kwargs):
         action.status = 'DONE'
         action.output = action_logs
         return action
+
+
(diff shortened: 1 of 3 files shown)


chalicelib/checks/audit_checks.py
```diff
@@ -782,3 +782,32 @@ def patch_bio_feature_organism_name(connection, **kwargs):
...
+    target_files = {}
+    for a_re in res:
+        if a_re.get('file_first_line'):
+            read_id = a_re['file_first_line'].split(' ')[0][1:]
+            if read_id.isnumeric():
+                target_files[a_re['@id']] = {'Experiment': a_re['experiments'][0]['@id'], 'Experiment_title': a_re['experiments'][0]['display_title']}
+
+    if target_files:
+        check.status = 'WARN'
+        check.brief_output = '%s fastq files have integer read ids' % len(target_files)
```
sbreiff, line 810: This could be check.summary or check.description rather than brief_output, it's not required to have a brief_output on every check and this is short enough to just display on the UI as summary or description without clicking an extra button.

chalicelib/checks/audit_checks.py
```diff
@@ -782,3 +782,32 @@ def patch_bio_feature_organism_name(connection, **kwargs):
...
+    query = '/search/?date_created.from=2020-04-13&file_format.file_format=fastq&status=uploaded&type=FileFastq'
+    res = ff_utils.search_metadata(query, key=connection.ff_keys)
+    if not res:
+        return check
+    target_files = {}
+    for a_re in res:
+        if a_re.get('file_first_line'):
+            read_id = a_re['file_first_line'].split(' ')[0][1:]
+            if read_id.isnumeric():
+                target_files[a_re['@id']] = {'Experiment': a_re['experiments'][0]['@id'], 'Experiment_title': a_re['experiments'][0]['display_title']}
```
sbreiff, line 806: For experiments with several/many fastq files, the experiment values here could be a bit redundant. Another option would be to create a new dict after this in which experiment is the key and the value is a dict containing a 'files' key:value pair - that way if there was only one new experiment uploaded with say 12 fastq files with int read_ids, the experiment appears once instead of 12 times. Both ways get the same information across but the way described here might be slightly more organized.

PR #270, opened 2020-04-23
Released HeLa check/action

andreacosolo:
Files containing sequences (fastq or bam) from HeLa cells should not be released (or other visible statuses).
Check 1: find released fastq and bam files.
Check 2: find released processed files (fastq or bam) deriving from workflows with restricted input files (and find also released fastq or bam processed files not deriving from any workflow).
Action (common for check1 and check2): set the status to 'restricted'.


chalicelib/checks/audit_checks.py
```diff
@@ -823,3 +823,124 @@ def check_fastq_read_id(connection, **kwargs):
...
+                    files['visible'].append({
+                        'uuid': a_file['value']['uuid'],
+                        'file_format': format,
+                        'file_status': status})
+    # search for visible fastq or bam processed files that are not output of any workflow
+    query_pf = 'search/?type=FileProcessed&workflow_run_outputs.workflow.title=No+value'
+    query_pf += ''.join(['&status=' + st for st in visible_statuses])
+    query_pf += ''.join(['&file_format.file_format=' + f for f in formats])
+    query_pf += '&field=uuid&field=status&field=file_format'
+    res_pf = ff_utils.search_metadata(query_pf, key=connection.ff_keys)
```
KorayKirli, line 897: This part is not clear for me, when does HeLa come into play?
If a submitter sends their own processed bam file for a non HeLa experiment set, will this check complain about that file?
In other words, if you were to not limit this check to bam and fastq, you would find many files.

andreacosolo: The part above this block looks for visible processed files (fastq or bam) that are output of a workflow, where the inputs are HeLa (or any other restricted file). It is important to ensure that there are no visible processed files (fastq or bam) 'unlinked' from workflows (as could happen with user-submitted pf), otherwise they will not be found by the search above.

The focus of this second check not strictly on HeLa, but rather on "Processed files not restricted". If some pf are not restricted but should be, they are returned as 'visible'. If some pf cannot be determined because the workflow info is missing, they are returned as 'unlinked'. The action patches only on the 'visible' ones, since the 'unlinked' ones require to be linked manually.

This part of the code (finding unlinked pf) can be a check itself or can be included in this check. I'm happy to discuss how useful it would be on its own, outside this specific application. Since it is essential for this check to work, but I do not see a strong need for a broader application, I thought of including it here. In particular, this search currently returns 0 files (apart from those few which you corrected earlier, that are not likely to show up again). So it seems unlikely that many of these files will start appearing all of a sudden. Moreover, extending the search to other file formats will give many (>2000) results. Only if we want to manually fix them all, an autonomous check would make sense.

KorayKirli: If you are concerned about processed results on HeLa experiment sets submitted by users, you can run a version of your fastq query for processed_files.

aschroed: I say leave this bit in - it's not going to hurt and it is potentially informative even if not hela specific.  If it starts turning up files that should rightfully be in this state - seems unlikely - then we can modify to add a white list in the future.  A complementary approach that might be done in a separate more general check would be to find a way to find orphan files by looking for those either not linked to experiment or experiment_set or not as output of a workflow or input of something other than md5 or fastqc (for fastq) to find truly orphaned files but don't think it is a priority.

I'd say this is good to go.

PR #271, opened 2020-04-24
C4-136 DELETE env API

willronchetti:
- Provides an API endpoint for deleting foursight environments

### DIFF
--- a/app.py
+++ b/app.py
@@ -296,7 +296,7 @@ def delete_environment(environ):
 
     Protected route
     """
-    if check_authorization(app.current_request.to_dict(), environ):
+    if check_authorization(app.current_request.to_dict(), environ):  # TODO (C4-138) Centralize authorization check
         return run_delete_environment(environ)
     else:
         return forbidden_response()
--- a/app.py
+++ b/app.py
@@ -286,6 +286,22 @@ def get_environment_route(environ):
     else:
         return forbidden_response()
 
+
+@app.route('/environments/{environ}/delete', methods=['DELETE'])
+def delete_environment(environ):
+    """
+    Takes a DELETE request and purges the foursight environment specified by 'environ'.
+    NOTE: This only de-schedules all checks, it does NOT wipe data associated with this
+    environment - that can only be done directly from S3 (for safety reasons).
+
+    Protected route
+    """
+    if check_authorization(app.current_request.to_dict(), environ):
+        return run_delete_environment(environ)
+    else:
+        return forbidden_response()
+
+
 ######### PURE LAMBDA FUNCTIONS #########
 
 @app.lambda_function()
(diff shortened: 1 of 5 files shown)


app.py
```diff
@@ -286,6 +286,22 @@ def get_environment_route(environ):
...
+@app.route('/environments/{environ}/delete', methods=['DELETE'])
+def delete_environment(environ):
+    """
+    Takes a DELETE request and purges the foursight environment specified by 'environ'.
+    NOTE: This only de-schedules all checks, it does NOT wipe data associated with this
+    environment - that can only be done directly from S3 (for safety reasons).
+
+    Protected route
+    """
+    if check_authorization(app.current_request.to_dict(), environ):
```
netsettler, line 299: I looked at the check_authorization function and while it looks probably-OK, I think it should not be maintained as part of foursight. We should centralize all authy things in a repo that is ultimately maintained by devops, and probably supplies decorators that are purely declarative for code so that we know things are being done in a standard way that has been vetted. I don't think that has to be done on this PR, but a # TODO (C4-138): comment would be nice. (That's the ticket number I just created.)

chalicelib/app_utils.py
```diff
@@ -771,6 +771,47 @@ def run_get_environment(environ):
...
+    but does not remove any data.
+    """
+    s3_connection = S3Connection('foursight-envs')
+    s3_resp = s3_connection.delete_keys([environ])
+    keys_deleted = s3_resp['Deleted']
+    if not keys_deleted:
+        response = Response(
+            body = {
+                'status': 'error',
+                'description': 'Invalid environment provided to delete',
```
netsettler, line 786: Is that really the only reason this could be? I'm not a big fan of descriptions that just make something up after-the-fact. I might say "Unable to comply with request."

chalicelib/app_utils.py
```diff
@@ -771,6 +771,47 @@ def run_get_environment(environ):
...
+            },
+            status_code = 400
+        )
+    else:
+        our_key = keys_deleted[0]  # since we only passed on key to be deleted, the response will be a length 1 list
+        if our_key['Key'] != environ:
+            response = Response(
+                body = {
+                    'status': 'error',
+                    'description': 'An error occured during environment deletion, please check S3 directly',
```
netsettler, line 797: typo: 'occurred' has two r's. 

But also, is there any security danger in revealing the nature of the error in a more clear way?
```
'description': 'An error occurred during environment deletion: the environment named in the response (%s) did not match the environment given (%s). Please check S3 directly' % (our_key['Key'], environ),
```
If the `Key` is potentially a private object, it's OK to leave out the parentheticals in the above and just say that the object returned didn't match.

Finally, is `our_key` really the right name for this result? At minimum you don't know if it's ours until you check it matches, but even beyond that, is the object out of which `'Key'` is taken itself just a structured representation of a key?

willronchetti: `our_key` is okay IMO because it should never be anything else except a length 1 list holding the key we requested. If it returns something else, that would mean a problem occurred on the AWS side (not due to us) - which we would like to catch if it does somehow happen.

PR #274, opened 2020-04-30
Pmid geo links

sbreiff:
`biorxiv_is_now_published` check edited to look for GEO datasets associated with a publication when new publications are found

### DIFF
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -143,7 +143,7 @@ def biorxiv_is_now_published(connection, **kwargs):
     chkstatus = ''
     chkdesc = ''
     check.action = "add_pub_and_replace_biorxiv"
-    fulloutput = {'biorxivs2check': {}, 'false_positives': {}}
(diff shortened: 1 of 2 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -243,6 +243,18 @@ def biorxiv_is_now_published(connection, **kwargs):
             # we have possible article(s) - populate check_result
             fndcnt += 1
             fulloutput['biorxivs2check'][buuid] = ['PMID:' + id for id in ids]
+            # look for GEO datasets
+            for id_ in ids:
+                result = requests.get('https://eutils.ncbi.nlm.nih.gov/entrez/eutils/'
```
aschroed, line 248: Should this be in a try-except block as it's ancillary and not sure check should error if the get doesn't work?

sbreiff: I  could do that - should I do the same for all gets? e.g. line 220 `res = requests.get(title_query)`?

aschroed: Yes - try-except is not the only option - just need some way to be sure that result.text is something you can actually work with

PR #276, opened 2020-05-07
dbxrefs formatting check

andreacosolo:


chalicelib/checks/wrangler_checks.py
```diff
@@ -1842,3 +1842,30 @@ def add_suggested_enum_values(connection, **kwargs):
...
+@check_function()
+def check_external_references_uri(connection, **kwargs):
+    '''
+    Check if external_references.uri is missing while external_references.ref
+    is present. Note that this check does not detect cases where multiple
+    references are present but not all have problems.
+    '''
+    check = CheckResult(connection, 'check_external_references_uri')
+    search = ('search/?type=Item&dbxrefs%21=No+value'
+              '&external_references.uri=No+value&field=dbxrefs&field=@id')
```
sbreiff, line 1856: `@id` is always returned, you don't need to specify `&field=@id`

chalicelib/checks/wrangler_checks.py
```diff
@@ -1842,3 +1842,30 @@ def add_suggested_enum_values(connection, **kwargs):
...
     return action
+
+
+@check_function()
+def check_external_references_uri(connection, **kwargs):
+    '''
+    Check if external_references.uri is missing while external_references.ref
+    is present. Note that this check does not detect cases where multiple
+    references are present but not all have problems.
+    '''
```
aschroed, line 1853: If we had nested mapping in fourfront it would be easy peasy to do this with a simple search.  I still think it might be worthwhile to try to check all cases, which I think we can do by only looking at the external_references property.  Something like:
search = "/search/?type=Item&external_references.ref%21=No+value&field=external_references"
result = ff_utils.search_metadata(search, key=connection.ff_keys)
for res in result:
    baddies = [er for er in res.get('external_references', []) if not er.get('uri')]
    if baddies:
             #report them

Parameters to search_metadata - eg. make it a generator with a page limit if needed.

PR #284, opened 2020-05-18
C4-58 Deployment Checks

willronchetti:
Adds the following checks/actions:
- `indexer_server_status`: will report the status of currently running indexer servers given a base environment. For example, if you spun up an indexer server for `fourfront-hotseat`, you would pass `fourfront-hotseat` as the `env` kwarg.
- `terminate_indexer_server`: action associated with `indexer_server_status`. Run the above check to get access to this action, which when run will terminate the associated indexer server. You can also force an indexer server environment with the `forced_env` kwarg.
- `provision_indexer_environment`: check that when queued will spin up an indexer server for the given environment and application_version, both required. To see the application version, navigate to the EB console and select the relevant environment - it is at the top in-between health and platform.
- `deploy_application_to_beanstalk`: check that will deploy the application to Elastic Beanstalk given an environment, branch, application_version (to use, optional) and repository (if you want a CGAP deployment, optional for FF). This check will allow us to very easily schedule deployments through Foursight for all environments as desired.

### DIFF
--- a/app.py
+++ b/app.py
@@ -13,6 +13,12 @@ DEFAULT_ENV = 'data'
 
 '''######### SCHEDULED FXNS #########'''
 
+
+def effectively_never():
+    """Every February 31st, a.k.a. 'never'."""
+    return Cron('0', '0', '31', '2', '?', '?')
+
+
 # this dictionary defines the CRON schedules for the dev and prod foursight
 # stagger them to reduce the load on Fourfront. Times are UTC
 # info: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html
@@ -28,7 +34,7 @@ foursight_cron_by_schedule = {
         'morning_checks_2': Cron('15', '10', '*', '*', '?', '*'),
         'monday_checks': Cron('0', '9', '?', '*', '2', '*'),
         'monthly_checks': Cron('0', '9', '1', '*', '?', '*'),
-        'manual_checks': Cron('0', '0', '31', '2', '?', '?')  # every February 31st aka never
+        'manual_checks': effectively_never()
     },
     'dev': {
         'ten_min_checks': Cron('5/10', '*', '*', '*', '?', '*'),
@@ -40,7 +46,7 @@ foursight_cron_by_schedule = {
         'morning_checks_2': Cron('45', '10', '*', '*', '?', '*'),
         'monday_checks': Cron('30', '9', '?', '*', '2', '*'),
         'monthly_checks': Cron('30', '9', '1', '*', '?', '*'),
-        'manual_checks': Cron('0', '0', '31', '2', '?', '?')
+        'manual_checks': effectively_never()
     }
 }
 
@@ -88,6 +94,11 @@ def monday_checks(event):
 @app.schedule(foursight_cron_by_schedule[STAGE]['monthly_checks'])
 def monthly_checks(event):
     queue_scheduled_checks('all', 'monthly_checks')
+
+
+@app.schedule(foursight_cron_by_schedule[STAGE]['manual_checks'])
+def manual_checks(event):  # should have no effect since cron schedule will never run
+    queue_scheduled_checks('all', 'manual_checks')
 
 
 '''######### END SCHEDULED FXNS #########'''
--- a/app.py
+++ b/app.py
@@ -27,7 +27,8 @@ foursight_cron_by_schedule = {
         'morning_checks': Cron('0', '10', '*', '*', '?', '*'),
         'morning_checks_2': Cron('15', '10', '*', '*', '?', '*'),
         'monday_checks': Cron('0', '9', '?', '*', '2', '*'),
-        'monthly_checks': Cron('0', '9', '1', '*', '?', '*')
+        'monthly_checks': Cron('0', '9', '1', '*', '?', '*'),
+        'manual_checks': Cron('0', '0', '31', '2', '?', '?')  # every February 31st aka never
     },
     'dev': {
         'ten_min_checks': Cron('5/10', '*', '*', '*', '?', '*'),
@@ -38,7 +39,8 @@ foursight_cron_by_schedule = {
         'morning_checks': Cron('30', '10', '*', '*', '?', '*'),
         'morning_checks_2': Cron('45', '10', '*', '*', '?', '*'),
         'monday_checks': Cron('30', '9', '?', '*', '2', '*'),
-        'monthly_checks': Cron('30', '9', '1', '*', '?', '*')
+        'monthly_checks': Cron('30', '9', '1', '*', '?', '*'),
+        'manual_checks': Cron('0', '0', '31', '2', '?', '?')
     }
 }
 
--- a/chalicelib/check_setup.json
+++ b/chalicelib/check_setup.json
@@ -1562,19 +1562,40 @@
     "indexer_server_status": {
         "title": "Indexer Server Status",
         "group": "Deployment Checks",
-        "schedule": {},
-        "display": ["data"]
+        "schedule": {
+            "manual_checks": {
+                "data": {
+                    "kwargs": {
+                        "primary": true
+                    }
+                }
+            }
+        }
     },
     "provision_indexer_environment": {
         "title": "Provision Indexer Environment",
         "group": "Deployment Checks",
-        "schedule": {},
-        "display": ["data"]
+        "schedule": {
+            "manual_checks": {
+                "data": {
+                    "kwargs": {
+                        "primary": true
+                    }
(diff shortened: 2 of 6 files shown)


app.py
```diff
@@ -27,7 +27,8 @@
         'morning_checks': Cron('0', '10', '*', '*', '?', '*'),
         'morning_checks_2': Cron('15', '10', '*', '*', '?', '*'),
         'monday_checks': Cron('0', '9', '?', '*', '2', '*'),
-        'monthly_checks': Cron('0', '9', '1', '*', '?', '*')
+        'monthly_checks': Cron('0', '9', '1', '*', '?', '*'),
+        'manual_checks': Cron('0', '0', '31', '2', '?', '?')  # every February 31st aka never
```
netsettler, line 31: Uhhh. Wow. In general I wish for these to be expressed in a more readable way, though we don't have to solve that today. But as to this one new need, how about...
```
def effectively_never():
    """Every February 31st, a.k.a. 'never'."""
    return Cron('0', '0', '31', '2', '?', '?')
```

app.py
```diff
@@ -38,7 +39,8 @@
         'morning_checks': Cron('30', '10', '*', '*', '?', '*'),
         'morning_checks_2': Cron('45', '10', '*', '*', '?', '*'),
         'monday_checks': Cron('30', '9', '?', '*', '2', '*'),
-        'monthly_checks': Cron('30', '9', '1', '*', '?', '*')
+        'monthly_checks': Cron('30', '9', '1', '*', '?', '*'),
+        'manual_checks': Cron('0', '0', '31', '2', '?', '?')
```
netsettler, line 43: Oh, look. Another use already. Perfect for having a common function. :)

chalicelib/check_setup.json
```diff
@@ -1558,5 +1558,44 @@
                 }
             }
         }
+    },
+    "indexer_server_status": {
```
netsettler, line 1562: Is this info not possible to arrange as part of a decorator? Why is it separated from the function name? (This doesn't have to be solved for this PR, just curious.)

chalicelib/checks/deployment_checks.py
```diff
@@ -0,0 +1,189 @@
...
+@check_function(env='fourfront-mastertest',
+                branch='master',
+                application_version_name=None, repo=None)
+def deploy_application_to_beanstalk(connection, **kwargs):
+    """ Deploys application to beanstalk under the given application_version name + a branch
+
+        NOTE: CGAP requires kwargs!
+    """
+    check = CheckResult(connection, 'deploy_application_to_beanstalk')
+    env = kwargs.get('env', 'fourfront-mastertest')  # by default
```
netsettler, line 159: What facility would set these to other than the default values?

chalicelib/checks/deployment_checks.py
```diff
@@ -0,0 +1,189 @@
...
+
+    if repo is not None:
+        repo_location = clone_repo_to_temporary_dir(repo)
+    else:
+        repo_location = clone_repo_to_temporary_dir()
+
+    packaging_was_successful = EBDeployer.build_application_version(repo_location,
+                                                                    application_version_name,
+                                                                    branch=branch)
+    time.sleep(10)  # give EB some time to index the new template
```
netsettler, line 175: Is there really no way to ask?

chalicelib/checks/deployment_checks.py
```diff
@@ -0,0 +1,189 @@
...
+        try:
+            EBDeployer.deploy_new_version(env, repo_location, application_version_name)
+            check.status = 'PASS'
+            check.summary = 'Successfully deployed version %s to env %s' % (application_version_name, env)
+        except Exception as e:
+            check.status = 'ERROR'
+            check.summary = 'Error encountered while deploying: %s' % str(e)
+    else:
+        check.status = 'ERROR'
+        check.summary = 'Could not package repository: %s' % packaging_was_successful
```
netsettler, line 186: If we got here, packaging_was_successful will just have None or the empty string, right? I guess it can't hurt to show but it's a pretty boring value. Is there no error message somewhere?

chalicelib/checks/deployment_checks.py
```diff
@@ -0,0 +1,189 @@
...
+
+    cleanup_tempdir(repo_location)
```
netsettler, line 188: Should this be in a finally clause?

chalicelib/checks/es_checks.py
```diff
@@ -1,4 +1,5 @@
 import time
+import datetime
```
netsettler, line 2: I see you newly importing this but I don't see where you used it.

chalicelib/checks/es_checks.py
```diff
@@ -91,4 +94,4 @@ def clean_s3_es_checks(connection, **kwargs):
     full_output['n_deleted_es'] = n_deleted_es
     check.status = 'DONE'
     check.full_output = full_output
-    return action
+    return check
```
netsettler, line 97: So you're returning a different value. What was `action` coming from before and why wasn't it needed? I don't even see it being assigned.

chalicelib/checks/es_checks.py
willronchetti: If you look at this file you'll see `datetime` is referenced yet it is not imported. I'm guessing when this is packaged it gets it some other way, as I believe the checks work. I added the import so it's more clear.

chalicelib/checks/es_checks.py
willronchetti: I don't know where `action` came from, I noticed it while looking at this file and fixed it. I presume it meant to be `check` always.

chalicelib/checks/deployment_checks.py
willronchetti: Yup, probably a good idea.

chalicelib/checks/deployment_checks.py
willronchetti: It depends what happens in `build_application_version`. It could raise an error, return False, or return the response to the boto3 API call: `create_application_version`. Each one informs a different situation at least, but probably not an ideal API to have on my part.

chalicelib/checks/deployment_checks.py
willronchetti: You might be able to use: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticbeanstalk.html#ElasticBeanstalk.Client.update_configuration_template but as far as I can tell there isn't really a great way. 10 seconds is probably too long but wanted to be extra safe (we have quite a long time on lambda, these checks should never come close to timing out).

chalicelib/checks/deployment_checks.py
willronchetti: `app.run_check_or_action(connection, <check_name>, {'env': 'fourfront-green'})`.
I think once a check has been run it's arguments in the decorator will show up on the UI.

chalicelib/check_setup.json
willronchetti: Yeah, this info is all required and packaged separately from the code. See the first few methods in `check_utils.py`.

app.py
willronchetti: Can do. Probably a good idea anyway in case some mechanism is introduced to detect such schedules.

chalicelib/checks/deployment_checks.py
aschroed: Correct on arguments showing up on UI - I'm not positive but I think including the arguments in the decorator obviates the need for having a default in the .get request eg. here: kwargs.get('branch', 'master')
Is it still true that you need to run a check locally with primary=True to get it to show up in the UI?  If so then believe I'm correct but worth a test.

chalicelib/checks/deployment_checks.py
```diff
@@ -0,0 +1,193 @@
...
+        time.sleep(10)  # give EB some time to index the new template
+        if packaging_was_successful:
+            try:
+                EBDeployer.deploy_new_version(env, repo_location, application_version_name)
+                check.status = 'PASS'
+                check.summary = 'Successfully deployed version %s to env %s' % (application_version_name, env)
+            except Exception as e:
+                check.status = 'ERROR'
+                check.summary = 'Exception thrown while deploying: %s' % str(e)
+        else:
```
aschroed, line 184: If this happens the thing you are printing is 'False' I think which is not all that useful.

willronchetti: Yeah, it doesn't look useful but does in fact tell me something about what happened. The API itself in utils I think should be improved. It's hard to write the API first then go implement what uses it, you notice a lot of things when you go to use it you didn't see when implementing it.

PR #287, opened 2020-05-20
fix kwarg type

andreacosolo:


chalicelib/checks/wrangler_checks.py
```diff
@@ -1874,11 +1879,11 @@ def check_external_references_uri(connection, **kwargs):
     if items:
         check.status = 'WARN'
         check.summary = 'external_references.uri is missing'
-        check.description = '%s items missing uri' % len(items)
+        check.description = '%s items %sare missing uri' % (len(items), from_text)
```
aschroed, line 1882: very minor styling thing that caught my eye.  If you remove the space from end of string in line 1866 then you can add a space between %s and are in this line and 1886 and it not only reads nicer but is less ambiguous that you mean a string substitution.  Can take or leave as you wish.

andreacosolo: I did it like that because it can also be empty string (line 1860) in case all items are checked.

aschroed: Got it!

PR #289, opened 2020-05-21
C4-99 Nightly Deploys

willronchetti:
- Configures Foursight to deploy staging on weekdays at 6pm.
- Could adapt this for any environments we'd like to deploy on a schedule.

### DIFF
--- a/chalicelib/checks/deployment_checks.py
+++ b/chalicelib/checks/deployment_checks.py
@@ -8,9 +8,10 @@ from git import Repo
 from ..run_result import CheckResult, ActionResult
 from ..utils import check_function, action_function
 from dcicutils.deployment_utils import EBDeployer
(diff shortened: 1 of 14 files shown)


chalicelib/checks/deployment_checks.py
```diff
@@ -191,3 +193,18 @@ def deploy_application_to_beanstalk(connection, **kwargs):
         cleanup_tempdir(repo_location)
 
     return check
+
+
+def who_is_ff_staging():
```
netsettler, line 198: I made [utils PR #91](https://github.com/4dn-dcic/utils/pull/91) so (when that goes in) you can do
```
from dcicutils.beanstalk_utils import compute_ff_stg_env
```
to get this functionality.

willronchetti: Pulled this in today.

PR #290, opened 2020-05-22
opf contributing_lab

andreacosolo:
Supplementary files (opf) can have a different lab than the lab who generated the raw data (lab of the Experiment or ExperimentSet).
If this happens, the exp/set lab needs to be added as contributing_lab to the opf.

### DIFF
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -1897,7 +1897,8 @@ def check_opf_lab_different_than_experiment(connection, **kwargs):
     experimental lab needs to be added to the opf (contributing lab).
     '''
     check = CheckResult(connection, 'check_opf_lab_different_than_experiment')
-
+    check.action = add_contributing_lab_opf
+    
     # check only recently modified files, to reduce the number of items
     days_back = kwargs.get('days_back')
     try:
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -1908,7 +1908,7 @@ def check_opf_lab_different_than_experiment(connection, **kwargs):
     else:
         date_now = datetime.datetime.now(datetime.timezone.utc)
         date_diff = datetime.timedelta(days=days_back)
-        from_date = datetime.datetime.strftime(date_now - date_diff, "%Y-%m-%d")
+        from_date = datetime.datetime.strftime(date_now - date_diff, "%Y-%m-%d+%H:%M")
         from_date_query = '&last_modified.date_modified.from=' + from_date
         from_text = 'modified from %s ' % from_date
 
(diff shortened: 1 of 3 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -1887,3 +1887,110 @@ def check_external_references_uri(connection, **kwargs):
...
+              '&track_and_facet_info.experiment_bucket%21=processed+file' +
+              '&field=experiment_sets&field=experiments' +
+              '&field=lab&field=contributing_labs' + from_date_query)
+    result = ff_utils.search_metadata(search, key=connection.ff_keys)
+
+    exp_set_uuids = []  # Exp or ExpSet uuid list
+    for res in result:
+        if res.get('experiments'):
+            if len(res['experiments']) != 1:
+                return check  # this should not happen
```
sbreiff, line 1926: is returning the check really the desired behavior here rather than flagging them in the output and returning all the proper results? If the check gets returned as this stage will it be obvious what the problem is?

andreacosolo: Good point. This is one of the parts that needed refinement. I initially wrote it this way based on the track_and_facet_info code, which does a similar thing. 
At the moment, no such file is found. I agree that the check should flag these cases and return all the proper results. However, I am not sure what to do next with these cases. If they will be manually patched, then probably they should also be added to a list of files to ignore?

chalicelib/checks/wrangler_checks.py
```diff
@@ -1887,3 +1887,110 @@ def check_external_references_uri(connection, **kwargs):
...
+            return check  # this should not happen
+        res['exp_set_uuid'] = exp_or_set['uuid']
+        if res['exp_set_uuid'] not in exp_set_uuids:
+            exp_set_uuids.append(res['exp_set_uuid'])
+
+    # get lab of Exp/ExpSet
+    result_exp_set = ff_utils.get_es_metadata(exp_set_uuids, sources=['uuid', 'properties.lab'], key=connection.ff_keys)
+    uuid_2_lab = {}  # map file uuid to Exp/Set lab
+    for item in result_exp_set:
+        uuid_2_lab[item['uuid']] = item['properties']['lab']
```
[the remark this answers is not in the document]
andreacosolo: Are you suggesting a get() instead? I did not consider that an experiment or expset could be missing the lab. Actually, isn't that required?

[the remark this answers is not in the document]
aschroed: No - sorry I should have been clearer - I just wanted to be sure that what is being returned here is indeed a uuid but I just checked the function in utils and see that indeed you will always get the uuid when source = properties.field so all good.

chalicelib/checks/wrangler_checks.py
sbreiff: in the output you could separate actionable results and flagged results, and write the action so it only uses the actionable results. There probably doesn't need to be an "ignore" list, unless it is possible that some files won't meet that qualifier by design.

PR #292, opened 2020-05-28
C4-171 Fix nested checks

willronchetti:
- Some deployment checks tried to take advantage of nested checks, which do not function correctly out of the box. The check function itself runs but crashes in the decorator code.
- I have implemented a workaround to effectively allow nested checks in a specific scenario, but this should be revisited in the future. See associated comment.

### DIFF
--- a/chalicelib/checks/deployment_checks.py
+++ b/chalicelib/checks/deployment_checks.py
@@ -150,13 +150,13 @@ def provision_indexer_environment(connection, **kwargs):
     return check
 
 
-@check_function(env='fourfront-mastertest',
-                branch='master',
-                application_version_name=None, repo=None)
-def deploy_application_to_beanstalk(connection, **kwargs):
-    """ Deploys application to beanstalk under the given application_version name + a branch
+def _deploy_application_to_beanstalk(connection, **kwargs):
+    """ Creates and returns a CheckResult for 'deploy_application_to_beanstalk'.
 
-        NOTE: CGAP requires kwargs!
+        NOTE: this does NOT have the @check_function decorator because I'd like
+        to "call this check" in other checks. The decorator will effectively cast
+        the resulting check to a dictionary, breaking nested checks. This could be
+        fixed by refactoring "store_formatted_result". - Will 05/28/2020
     """
     check = CheckResult(connection, 'deploy_application_to_beanstalk')
     env = kwargs.get('env', 'fourfront-mastertest')  # by default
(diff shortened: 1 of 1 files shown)


chalicelib/checks/deployment_checks.py
```diff
@@ -150,13 +150,13 @@ def _deploy_indexer(e, version):
     return check
 
 
-@check_function(env='fourfront-mastertest',
-                branch='master',
-                application_version_name=None, repo=None)
-def deploy_application_to_beanstalk(connection, **kwargs):
-    """ Deploys application to beanstalk under the given application_version name + a branch
+def _deploy_application_to_beanstalk(connection, **kwargs):
```
netsettler, line 153: Sometime I would like to see the whole clone_repo_to_temporary_dir separated out to a context manager so you do:

with temporarily_cloned_repo(...)

It'd be fine for now to have a TODO.   But I find it tedious to have these things with multiple nested try situations all dancing around one another when really there are separate namable activities.

I also found myself wondering if this function would make sense to actually be part of a subclass of CheckResult if you do a lot of this.  Certainly the second function looks like more could be factored out. For example (untested):
```
def _deploy_repo_to_env(connection, repo, env, branch='master', **kwargs):
    if kwargs:  # Or maybe just ignore kwargs quietly?
        raise RuntimeError("Unknown kwargs: {kwargs}".format(kwargs=kwargs))
    this_check = CheckResult(connection, 'deploy_{repo}_{env}'.format(repo=repo, env=env) # e.g., 'deploy_fourfront_staging'
    helper_check = _deploy_application_to_beanstalk(connection, env=env, branch=branch)
    if helper_check.status == 'PASS':
        this_check.status = 'PASS'
        this_check.summary = 'Successfully deployed {repo} {branch} to {env}'.format(repo=repo, branch=branch, env=env)
    else:
        this_check.status = 'ERROR'
        this_check.summary = 'Error occurred during deployment, see full_output'
        this_check.full_output = helper_check.summary  # should have error message
    return this_check

@check_function
def deploy_ff_staging(connection, branch='master', **kwargs):
    return _deploy_repo_to_env(connection, repo='fourfront', env=compute_ff_stg_env(), branch=branch, **kwargs)
```
Though I still feel like all those status checks are still done twice when maybe they don't need to be. Perhaps a context manager would be a better way to encapsulate this behavior.

willronchetti: I've added a note for this.

PR #301, opened 2020-06-18
Upd biorxiv chk

aschroed:
Updated the check that looks for biorxivs that are now published in journals:

- changed the fulloutput data structure to include the data fields that will be transferred to the pub
- added a item for the biorxiv link so it can be used to populate the aka field
- modified the action to use the info from the check result during the transfer as possible

### DIFF
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -162,6 +162,10 @@ def biorxiv_is_now_published(connection, **kwargs):
         in the action - a parameter is provided to manually input a mapping between biorxiv (uuid)
         to journal article (PMID:ID) - to add that pairing to the result full_output. It will
         be acted on by the associated action format of input is uuid PMID:nnnnnn, uuid PMID:nnnnnn
+
+        NOTE: because the data to transfer from biorxiv to pub is not obtained from the check result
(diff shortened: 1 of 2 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -134,10 +158,14 @@ def biorxiv_is_now_published(connection, **kwargs):
 
         There are some examples of the title and author list being different enough so
         that the pubmid esearch query doesn't find the journal article.  In order to
-        allow the replacement, movement of all the relevant fields and adding static sections
+        allow the replacement, movement of all the relevant fields and adding replacement static sections
         in the action - a parameter is provided to manually input a mapping between biorxiv (uuid)
         to journal article (PMID:ID) - to add that pairing to the result full_output. It will
         be acted on by the associated action format of input is uuid PMID:nnnnnn, uuid PMID:nnnnnn
+
+        NOTE: because the data to transfer from biorxiv to pub is not obtained from the check result
```
andreacosolo, line 166: This is confusing to me, why do you say "not"? I thought the check result had the data to transfer in the 'data2transfer' field.

aschroed: Typo - meant 'now' but it actually makes more sense to remove the word entirely.

PR #305, opened 2020-07-01
add cgap-portal repo to cgap deploy function

ebberg:


chalicelib/checks/deployment_checks.py
```diff
@@ -259,4 +259,5 @@ def deploy_cgap_production(connection, **kwargs):
         env_to_deploy=compute_cgap_prd_env(),
         application_name="CGAP Portal",
         check='deploy_cgap_production',
+        repo='https://github.com/4dn-dcic/cgap-portal.git',
```
netsettler, line 262: We might want to think about having `deploy_env` allow just a `repo` name or about having a function `repo_url(...)` function so that typos in the URL are less likely and one can just focus on the name.

Case in point, the URL for `cgap` is: `https://github.com/dbmi-bgm/cgap-portal.git`

We have two areas, one for 4DN and one for CGAP. CGAP urls in the `dbmi-bgm` space.

PR #306, opened 2020-07-08
Higlass registration with local beddb files

alexander-veit:
During file registration on the Higlass server, `beddb` files will now be copied from S3 to the local file system of the Higlass server. A simple flask server component on the server is triggered via foursight that will copy the corresponding files. This was necessary, due to the low performance of mounted `beddb` files.

### DIFF
--- a/chalicelib/checks/higlass_checks.py
+++ b/chalicelib/checks/higlass_checks.py
@@ -1821,7 +1821,9 @@ def patch_file_higlass_uid(connection, **kwargs):
         'patch_failure': {},
         'patch_success': {},
         'registration_failure': {},
-        'registration_success': 0
+        'registration_success': 0,
+        'beddb_copy_failure': {},
+        'beddb_copy_success': 0
     }
     # get latest results
     higlass_check = CheckResult(connection, 'files_not_registered_with_higlass')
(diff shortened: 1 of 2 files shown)


chalicelib/checks/higlass_checks.py
```diff
@@ -1895,6 +1899,25 @@ def patch_file_higlass_uid(connection, **kwargs):
                 action_logs['registration_failure'][hit['accession']] = err_msg
                 continue
 
+            # Call the Flask server component on the Higlass server to copy beddb files
+            # from S3 to the local file system. Using mounted versions of these files is very slow. 
+            if payload['filetype'] == 'beddb':
+                copy_res = requests.post(
```
willronchetti, line 1905: No retry mechanism here? Might want to consider adding. Could be a pain to clean up if this does occasionally fail (not saying it will). If it does though it may be useful to have the ability to re-queue certain beddb files that fail, but if you've tested this and haven't seen it happen at all it may not be necessary.

alexander-veit: Thanks! I tested it and it seems to work fine so far. Whenever it failed, there was a larger issue (like permissions), so a retry wouldn't have helped anyway. 
Also, on initial registration, we can see if there was a problem, as it logs the copy failures. If it happens later and the file is not copied correctly, it will be automatically served from the S3 bucket.

PR #310, opened 2020-07-16
search url check

andreacosolo:


aschroed:
Assuming you have tested this looks good to me.

chalicelib/checks/audit_checks.py
```diff
@@ -349,6 +349,7 @@ def check_help_page_urls(connection, **kwargs):
                 # test only link part of match (not text part, even if it looks like a link)
                 idx = link.index(']')
                 url = link[link.index('(', idx)+1:-1]
+                urls.append(url)
```
sbreiff, line 352: Thanks for adding this, could you also change line 338 (`results = ...`) to use `&q=help&q=resources` instead of just `&q=help` ?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "foursight"
-version = "0.11.10"
+version = "0.11.11"
```
sbreiff, line 3: I'd say that since it isn't just a bug fix but adds a new check we could increment the minor version, e.g. 0.12.0 . But if others disagree i don't mind keeping it this way.

aschroed: I would agree with that - we should definitely try to get in the habit of minor version increments when functionality is added.

chalicelib/checks/audit_checks.py
```diff
@@ -378,6 +379,42 @@ def check_help_page_urls(connection, **kwargs):
...
+    for result in results:
+        body = result.get('body', '')
+        # search links for search or browse pages, either explicit or relative
+        urls = re.findall(r'[\(|\[|=]["]*(?:[^\s\)\]]+(?:4dnucleome|elasticbeanstalk)[^\s\)\]]+|/)((?:browse|search)/\?[^\s\)\]]+)[\)|\]|"]', body)
+        if urls:
+            for url in urls:
+                url = re.sub(r'&limit=[^&]*|limit=[^&]*&?', '', url)  # remove limit if present
+                q_results = ff_utils.search_metadata(url + '&limit=1&field=@id', key=connection.ff_keys)
+                if len(q_results) == 0:
+                    problematic_sections[result['@id']] = problematic_sections.get(result['@id'], [])
```
sbreiff, line 402: not something you need to change but just a future tip - you can also use `dict_name.setdefault(key_name, value)` if you need to check whether something is in the dictionary and create a new key if not.

PR #314, opened 2020-08-05
New check for "grouped with" file relation consistency

andreacosolo:
Check if "grouped with" file relationships are reciprocal and complete.
While other types of file relationships are automatically updated on the related file, "grouped with" ones need to be explicitly (manually) patched on the related file. This check ensures that there are no related files that lack the reciprocal relationship, or that lack some of the group relationships (for groups larger than 2 files).

### DIFF
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -2063,7 +2063,9 @@ def grouped_with_file_relation_consistency(connection, **kwargs):
     groups = []
     newgroups = [set(rel).union({file}) for file, rel in file2grp.items()]
 
-    # merge intersecting groups until there is any
+    # Check if any pair of groups in the list has a common file (intersection).
+    # In that case, they are parts of the same group: merge them.
+    # Repeat until all groups are disjoint (not intersecting).
     while len(groups) != len(newgroups):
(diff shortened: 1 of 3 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -2032,3 +2032,93 @@ def add_contributing_lab_opf(connection, **kwargs):
...
+            file2all.setdefault(f['@id'], []).append(
+                {"relationship_type": rel_type, "file": rel_file})
+            if rel_type == "grouped with":
+                file2grp.setdefault(f['@id'], []).append(rel_file)
+
+    # list groups of related items
+    groups = []
+    newgroups = [set(rel).union({file}) for file, rel in file2grp.items()]
+
+    # merge intersecting groups until there is any
```
aschroed, line 2066: Not sure what you mean exactly by this comment.

andreacosolo: Given the pairwise nature of file relationships, groups can be partially defined on each file. 
An initial list of (partial) groups is compiled by reading the relationships of each file. Now, if two partial groups have one or more common items, they need to be merged into one group. This needs to be repeated recursively on the list of partial groups, until there is no intersection.

PR #324, opened 2020-08-25
Refactor wipe_build_indices

willronchetti:
- Refactor `wipe_build_indices` to be FF/CGAP specific and run on the production schedule (morning checks) of each
- Instead of deleting `*`, we delete indices that have numeric prefixes. This convention has been communicated to others, but I don't think anything except Travis is using numerical prefixes.

### DIFF
--- a/chalicelib/checks/system_checks.py
+++ b/chalicelib/checks/system_checks.py
@@ -1,10 +1,9 @@
 from __future__ import print_function, unicode_literals
 from ..utils import (
     check_function,
-    action_function,
     basestring
 )
-from ..run_result import CheckResult, ActionResult
+from ..run_result import CheckResult
 from dcicutils import (
     ff_utils,
     es_utils,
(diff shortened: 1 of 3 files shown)


chalicelib/checks/system_checks.py
```diff
@@ -13,29 +12,51 @@
...
+# XXX: put into utils?
+CGAP_TEST_CLUSTER = 'search-cgap-testing-ud3ggpjj7x6vclx62nmzymyzfi.us-east-1.es.amazonaws.com:80'
+FF_TEST_CLUSTER = 'search-fourfront-testing-hjozudm7pq5wwlssx2pjlsyz7y.us-east-1.es.amazonaws.com:80'
+TEST_ES_CLUSTERS = [
+    CGAP_TEST_CLUSTER,
+    FF_TEST_CLUSTER
+]
+
+
+def wipe_build_indices(connection, es_url):
```
netsettler, line 30: The argument is not really a `URL`.  I would call it an `es_cluster` or `es_server` or `es_host` or `es_host_and_port`. Although maybe it's elsewhere called a URL and you're just going for consistency...Is there an `ES_URL` config param? Alas. Whatever you think is reasonable. I just wanted to mention the issue.

willronchetti: I chose `es_url` to be consistent with: https://github.com/4dn-dcic/utils/blob/master/dcicutils/es_utils.py#L11 but it is not a "standard" so to speak.

PR #334, opened 2020-09-10
Foursight ES6 Updates

willronchetti:
- Hook in new dcicutils, which will properly resolve ES if an HTTPS connection is requested.


netsettler (commented):
Seems fine.

PR #337, opened 2020-09-10
Check unique opfs

andreacosolo:
Check if opfs in an ExpSet are also raw, processed or reference files

### DIFF
--- a/chalicelib/checks/audit_checks.py
+++ b/chalicelib/checks/audit_checks.py
@@ -199,7 +199,7 @@ def expset_opf_unique_files(connection, **kwargs):
     '''
     check = CheckResult(connection, 'expset_opf_unique_files')
     days_back = kwargs.get('days_back')
-    from_date_query, from_text = last_modified_from(days_back)
+    from_date_query, from_text = wrangler_utils.last_modified_from(days_back)
     # opfs can be on Exps or ExpSets: search ExpSets for each case and merge results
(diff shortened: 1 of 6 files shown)


chalicelib/checks/audit_checks.py
```diff
@@ -190,6 +191,66 @@ def expset_opf_unique_files_in_experiments(connection, **kwargs):
...
+        all_files = []  # list all raw, processed and reference files in the ExpSet
+        if expset.get('other_processed_files'):
+            opfs_id.extend([f['@id'] for grp in expset['other_processed_files'] for f in grp.get('files', [])])
+        for exp in expset['experiments_in_set']:
+            if exp.get('other_processed_files'):
+                opfs_id.extend([f['@id'] for grp in exp['other_processed_files'] for f in grp.get('files', [])])
+            all_files.extend([f['@id'] for f in exp.get('files', [])])
+            all_files.extend([f['@id'] for f in exp.get('processed_files', [])])
+            all_files.extend([f['@id'] for f in exp.get('reference_files', [])])
+        all_files.extend([f['@id'] for f in expset.get('processed_files', [])])
```
sbreiff, line 233: if you're using `field=experiments_in_set&field=other_processed_files` in the search query, then it seems that processed_files won't be in the search result, no?

andreacosolo: Good catch! Fixed now

sbreiff (approved):

PR #341, opened 2020-09-14
Small fixes/optimizations

willronchetti:
- Use HTTPS for favicon always
- Use get_size instead of json.dumps, which should be faster
- Disable view permissions on CGAP envs for not-logged in users

When not logged in: https://kpqxwgx646.execute-api.us-east-1.amazonaws.com/api/view/cgapdev is now indistinguishable from https://kpqxwgx646.execute-api.us-east-1.amazonaws.com/api/view/blahblah

### DIFF
--- a/chalicelib/app_utils.py
+++ b/chalicelib/app_utils.py
@@ -376,6 +376,8 @@ def view_foursight(environ, is_admin=False, domain="", context="/"):
     view_envs = environments.keys() if environ == 'all' else [e.strip() for e in environ.split(',')]
     for this_environ in view_envs:
         try:
+            if 'cgap' in this_environ and not is_admin:  # no view permissions for non-admins on CGAP
+                continue
             connection = init_connection(this_environ, _environments=environments)
         except Exception:
             connection = None
--- a/chalicelib/app_utils.py
+++ b/chalicelib/app_utils.py
@@ -374,7 +374,6 @@ def view_foursight(environ, is_admin=False, domain="", context="/"):
     total_envs = []
     servers = []
     view_envs = environments.keys() if environ == 'all' else [e.strip() for e in environ.split(',')]
-    import pdb; pdb.set_trace()
     for this_environ in view_envs:
         try:
             connection = init_connection(this_environ, _environments=environments)
--- a/chalicelib/app_utils.py
+++ b/chalicelib/app_utils.py
@@ -14,6 +14,7 @@ from itertools import chain
 from dateutil import tz
 from base64 import b64decode
 from dcicutils import ff_utils
+from dcicutils.env_utils import is_cgap_env, is_cgap_server
 from .fs_connection import FSConnection
 from .run_result import CheckResult, ActionResult
 from .check_utils import (
@@ -42,6 +43,7 @@ from .s3_connection import S3Connection
 
 FOURFRONT_FAVICON = 'https://data.4dnucleome.org/static/img/favicon-fs.ico'
 CGAP_FAVICON = 'https://cgap.hms.harvard.edu/static/img/favicon-fs.ico'
+LAMBDA_MAX_BODY_SIZE = 5500000  # 6Mb is the "real" threshold
 
 jin_env = Environment(
     loader=FileSystemLoader('chalicelib/templates'),
@@ -184,7 +186,7 @@ def get_favicon(server):
         res = requests.head(favicon)
         return res.status_code == 200
 
-    if 'cgap' in server:
+    if is_cgap_server(server):
         favicon = CGAP_FAVICON  # want full HTTPS, so hard-coded in
     else:
         favicon = FOURFRONT_FAVICON  # *should* always be valid, so reasonable fallback
@@ -212,7 +214,7 @@ def process_response(response):
     Does any final processing of a Foursight response before returning it. Right now, this includes:
     * Changing the response body if it is greater than 5.5 MB (Lambda body max is 6 MB)
     """
-    if get_size(response.body) > 5500000:  # should be much faster than json.dumps
+    if get_size(response.body) > LAMBDA_MAX_BODY_SIZE:  # should be much faster than json.dumps
         response.body = 'Body size exceeded 6 MB maximum.'
         response.status_code = 413
     return response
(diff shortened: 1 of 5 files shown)


chalicelib/app_utils.py
```diff
@@ -179,13 +180,15 @@ def get_favicon(server):
...
-        favicon = server + 'static/img/favicon-fs.ico'
+    def favicon_is_valid(favicon):
         res = requests.head(favicon)
-        assert res.status_code == 200
-    except:
-        favicon = DEFAULT_FAVICON
-    return favicon
+        return res.status_code == 200
+
+    if 'cgap' in server:
```
netsettler, line 187: There's a better way to do this if you took the bs_env as an argument, using `dcicutils.env_utils.is_cgap_env`. But even if not, we should maybe have a similar `is_cgap_server(server)` or even `get_favicon_url(bs_env)` rather than do string operations here.

chalicelib/app_utils.py
```diff
@@ -373,6 +376,8 @@ def view_foursight(environ, is_admin=False, domain="", context="/"):
     view_envs = environments.keys() if environ == 'all' else [e.strip() for e in environ.split(',')]
     for this_environ in view_envs:
         try:
+            if 'cgap' in this_environ and not is_admin:  # no view permissions for non-admins on CGAP
```
netsettler, line 379: definitely `is_cgap_env` here.

chalicelib/app_utils.py
```diff
@@ -209,7 +212,7 @@ def process_response(response):
     Does any final processing of a Foursight response before returning it. Right now, this includes:
     * Changing the response body if it is greater than 5.5 MB (Lambda body max is 6 MB)
     """
-    if len(json.dumps(response.body)) > 5500000:
+    if get_size(response.body) > 5500000:  # should be much faster than json.dumps
```
netsettler, line 215: I'm not sure if get_size is a builtin of some sort or something we wrote. If a builtin, I might still write a separate function (untested):
```
def response_body_size_exceeds(response, max_size):
    return get_size(response.body) > max_size  # should be faster than measuring against json.dumps
```
so you can separate heuristics from policy. Then (untested):
```
HUGE_RESPONSE_BODY_THRESHOLD = 5500000

def process_response(response):
    if response_body_size_exceeds(response, HUGE_RESPONSE_BODY_THRESHOLD):
        response.body = ("Body size exceeded %s MB maximum."
                         % HUGE_RESPONSE_BODY_THRESHOLD / 1.0e6)
        response.status_code = 413
    return response
```
Although I might have introduced other bugs along the way, this would among other things make the function not lie about the size it's looking at, which is 5.5 MB, not 6 MB, though if you wanted to blur it you could still make the code less fragile by just doing the division and rounding up. The real point here is to insulate you against the message being wrong if the threshold number is changed and someone doesn't notice that the string is trying to be sync'd with that value.

This modularity also means you could choose to optimize the response body calculation to stop after it knew it was bigger than the threshold amount.

chalicelib/app_utils.py
```diff
@@ -403,7 +408,7 @@ def view_foursight(environ, is_admin=False, domain="", context="/"):
     queue_attr = get_sqs_attributes(get_sqs_queue().url)
     running_checks = queue_attr.get('ApproximateNumberOfMessagesNotVisible')
     queued_checks = queue_attr.get('ApproximateNumberOfMessages')
-    first_env_favicon = get_favicon(servers[0]) # use first env
+    first_env_favicon = get_favicon(servers[0]) if len(servers) > 0 else get_favicon('ff')  # use first env, default FF
```
netsettler, line 411: Could you add a comment saying why "first" is relevant here? I'm out of context so don't know if both cgap and ff servers could be in play or whether it's just that you know all servers are cgap or all are ff.

chalicelib/app_utils.py
```diff
@@ -602,7 +607,7 @@ def view_foursight_history(environ, check, start=0, limit=25, is_admin=False,
     if server is not None:
         favicon = get_favicon(server)
     else:
-        favicon = DEFAULT_FAVICON
+        favicon = FOURFRONT_FAVICON
```
netsettler, line 610: Even if FOURFRONT is the right default, why is the variable name not `DEFAULT_FAVICON` or at least `DEFAULT_DEFAULT_FAVICON` (presumably for when ecosystem-based defaulting has failed). Though this is often a place where having a third icon that is the "I couldn't find one" icon makes bugs get fixed quicker.

Anyway, this function looks suspiciously Foursight-centric in a set of changes that seem to be going the other way.

chalicelib/fs_connection.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, fs_environ, fs_environ_info, test=False, use_es=True):
         self.ff_es = fs_environ_info['es']
         if not test:
             self.ff_s3 = s3Utils(env=self.ff_env)
-            try:
+            try:  # TODO: make this configurable from env variables?
```
netsettler, line 41: Env variables are useful for bootstrapping the system, but aren't those variable supposed to get shuttled into the registry for use in the application? It's possible that security keys should get copied into aws credential files and not the registry...

tests/test_app_utils.py
```diff
@@ -24,7 +24,7 @@ def test_init_connection(self):
 
     def test_get_favicon(self):
         """ Tests that given 'mastertest' we get the right url for favicon """
-        expected = self.conn.ff_server + 'static/img/favicon-fs.ico'
+        expected = 'https://data.4dnucleome.org/static/img/favicon-fs.ico'
```
netsettler, line 27: Possibly of help to avoid wired strings:
```
>>> from dcicutils.env_utils import public_url_mappings
>>> public_url_mappings('fourfront-cgap')
{'cgap': 'https://cgap.hms.harvard.edu', 'data': 'https://cgap.hms.harvard.edu', 'staging': 'https://staging.cgap.hms.harvard.edu'}
>>> public_url_mappings('fourfront-mastertest')
{'staging': 'http://staging.4dnucleome.org', 'data': 'https://data.4dnucleome.org'}
```

netsettler (commented):
I'll just leave these as comments for now, holding back on approval pending discussion.

chalicelib/app_utils.py
willronchetti: When navigating to foursight it is legal to navigate to http://foursight.4dnucleome.org/view/data,cgap for example at which point for favicon we just select the first one in the specified list.

netsettler (approved):
This looks OK.

PR #350, opened 2020-09-30
Correct bugs in item mismatch and opf mismatch checks

aschroed:
- Updated STATUS_LEVEL scores to allow released to project to be distinguished from released in both item_mismatch checks.
- Set 'archived...' statuses to equivalent scores to 'released...' statuses.
- Added a condition to make sure the levels were the same when there are more than one status found
- Added a condition to ignore empty opf collections that was causing an exception to be thrown.

### DIFF
--- a/chalicelib/checks/audit_checks.py
+++ b/chalicelib/checks/audit_checks.py
@@ -676,8 +676,7 @@ def check_opf_status_mismatch(connection, **kwargs):
                 if not (list(statuses)[0] == 'pre-release' and opf_status_dict[hg_dict[title]] == 'released to lab'):
                     problem_dict[title] = {'files': list(statuses)[0],
                                            'higlass_view_config': opf_status_dict[hg_dict[title]]}
-            elif 'released' not in result['status'] and (STATUS_LEVEL[result['status']] < STATUS_LEVEL[list(statuses)[0]]):
-                # if ExpSet not released, and opf collection has higher status
+            elif STATUS_LEVEL[result['status']] < STATUS_LEVEL[list(statuses)[0]]:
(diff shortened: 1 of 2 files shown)


chalicelib/checks/audit_checks.py
```diff
@@ -662,15 +662,21 @@ def check_opf_status_mismatch(connection, **kwargs):
...
+                if len(scores) > 1:
+                    problem_dict[title] = {f['@id']: {'status': opf_status_dict[f['uuid']]} for f in file_list}
+                    if hg_dict.get(title):
+                        problem_dict[title][hg_dict[title]] = {'status': opf_status_dict[hg_dict[title]]}
             elif hg_dict.get(title) and STATUS_LEVEL[list(statuses)[0]] != STATUS_LEVEL[opf_status_dict[hg_dict[title]]]:
                 if not (list(statuses)[0] == 'pre-release' and opf_status_dict[hg_dict[title]] == 'released to lab'):
                     problem_dict[title] = {'files': list(statuses)[0],
                                            'higlass_view_config': opf_status_dict[hg_dict[title]]}
-            elif 'release' not in result['status'] and (STATUS_LEVEL[result['status']] < STATUS_LEVEL[list(statuses)[0]]):
+            elif 'released' not in result['status'] and (STATUS_LEVEL[result['status']] < STATUS_LEVEL[list(statuses)[0]]):
```
andreacosolo, line 679: If the ExpSet is "released to project" and opfs are "released", this will not be caught, right?

aschroed: Good catch.  I entirely removed the 'released' bit and currently getting same result on data.  Maybe something will show up for a submission that is in progress that will remind us why that was there but I think it should work.

andreacosolo (approved):

PR #366, opened 2020-11-12
Tweak sched

aschroed:
Updated cron schedules to 4 different morning_checks and 3 different hourly_checks and modified check_setup.json to use them.


chalicelib/check_setup.json
```diff
@@ -487,1247 +597,1556 @@
...
+    "check_status_mismatch": {
+        "title": "Experiment Sets containing Items with mismatch status",
+        "group": "Audit checks",
         "schedule": {
-            "morning_checks": {
-                "mastertest": {
-                    "kwargs": {"primary": true},
-                    "dependencies": []
+            "morning_checks_2": {
+                "all_4dn": {
```
willronchetti, line 613: We should be careful about scheduling checks this way, is there necessarily a need? We should consider a concept of "suppressed" schedules for (Chalice) dev and (ElasticBeanstalk) test environments, particularly for checks that run very frequently. This will save us resources over time.

chalicelib/checks/system_checks.py
```diff
@@ -99,72 +99,72 @@ def elastic_search_space(connection, **kwargs):
...
-        check.status = 'FAIL'
-        check.summary = check.description = 'EB environment is degraded; requests are likely to fail'
-    else:
-        check.summary = check.description = 'EB environment seems healthy'
-        check.status = 'PASS'
-    check.full_output = full_output
-    return check
+# @check_function()
+# def elastic_beanstalk_health(connection, **kwargs):
+#     """
```
willronchetti, line 104: You can just comment out the decorator - no need to comment out the entire thing, similar to how you did `data_release_updates` for example.

willronchetti (approved):

PR #370, opened 2020-12-01
external_submission dbxref check

andreacosolo:
Check if items with external_submission also have dbxrefs.


aschroed (approved):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "foursight"
-version = "1.3.1"
+version = "1.4.1"
```
sbreiff, line 3: should be 1.4.0

andreacosolo: good catch! thanks

chalicelib/checks/audit_checks.py
```diff
@@ -1074,3 +1075,37 @@ def restrict_hela(connection, **kwargs):
...
+    if items:
+        check.status = 'WARN'
+        check.summary = 'Items missing dbxrefs found'
+        check.description = '{} items exported for external submission more than {} days ago but still without dbxrefs'.format(len(items), delay)
+
+    else:
+        check.status = 'PASS'
+        check.summary = 'No items missing dbxrefs found'
+        check.description = 'All items exported for external submission more than {} days ago have dbxrefs'.format(delay)
+    check.full_output = [(i['@id'], i['external_submission']) for i in items]
```
sbreiff, line 1109: this might display better as a dict (e.g. `{i['@id']: i['external_submission'] for i in items}` but if you find it easier to use this way then I don't have a problem with it.

chalicelib/checks/audit_checks.py
```diff
@@ -1074,3 +1075,37 @@ def restrict_hela(connection, **kwargs):
...
+        delay = int(delay)
+    except (ValueError, TypeError):
+        delay = 14
+    date_now = datetime.datetime.now(datetime.timezone.utc)
+    days_diff = datetime.timedelta(days=delay)
+    to_date = datetime.datetime.strftime(date_now - days_diff, "%Y-%m-%d %H:%M")
+
+    query = ('search/?type=Item&dbxrefs=No+value' +
+             '&external_submission.date_exported.to=' + to_date)
+    items = ff_utils.search_metadata(query + '&field=dbxrefs&field=external_submission', key=connection.ff_keys)
```
sbreiff, line 1099: adding `&frame=object` here would make this a little quicker.

sbreiff (approved):
I added a few comments but nothing blocking.

chalicelib/checks/audit_checks.py
andreacosolo: Since I restricted the query to only those two fields, I thought that changing the frame would not make any difference. Anyway, I don't expect this search to return too many items, in theory.

sbreiff: there's a difference but it's small.

PR #384, opened 2021-01-15
User oh sync

Lmercadom:


aschroed (commented):
I think we need to clarify with OH are they maintaining phase I users in their list - i.e, those that are still part of phase I labs so maybe should not be made completely inactive?  And I think before we can implement this in anger we also need to get clarification from the consortium on if phase I users can be left with access to new project data that is not yet released or if we need to strictly control this.  Depending on the answers we will likely need to make some modifications to the viewing_groups properties based on their phase 1 or 2 ness and perhaps add explicit tags to awards.  I will try to get some clarification on this stuff.  In the meantime I've left some comments.  I think there may be other places where things could possibly be made more efficient but overall I think it looks like this will do the trick if some of this gets addressed.  It would be good to start getting some useful data from OH though - presumably once they get things updated dumping the data to the sheets should be quick.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2270,3 +2276,480 @@ def patch_hic_summary_tables(connection, **kwargs):
...
+                    add_awards = [i['uuid'] for i in all_grants if a_record['OH Grant'] in i['@id']]
+                    if add_awards:
+                        add_award = add_awards[0]
+                    else:
+                        add_award = a_record['OH Grant']
+                    problem.append(['cannot find the lab', a_record['OH Lab'], add_award])
+                    continue
+                actions['add_user'].append(user_data)
+
+    all_patching_uuids = [v['DCIC UUID'] for k, v in actions['patch_excel'].items() if v.get('DCIC UUID')]
```
aschroed, line 2578: can use .values() rather than .items()

chalicelib/checks/wrangler_checks.py
```diff
@@ -2270,3 +2276,480 @@ def patch_hic_summary_tables(connection, **kwargs):
...
+    # skip the total
+    skip_list = all_dcic_uuids + all_patching_uuids + skip_users_uuid
+    remaining_users = [i for i in fdn_users if i['uuid'] not in skip_list]
+    print(len(remaining_users), 'remaining users')
```
aschroed, line 2582: Same comment on print.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2270,3 +2276,480 @@ def patch_hic_summary_tables(connection, **kwargs):
...
+        check.status = 'WARN'
+        check.summary += '| %s problems' % (str(len(problem)))
+
+    check.full_output = {'actions': actions, 'problems': problem}
+    return check
+
+
+@action_function()
+def sync_users_oh_start(connection, **kwargs):
+    """Start runs by sending compiled input_json to run_workflow endpoint"""
```
aschroed, line 2609: Remove or update comment.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2270,3 +2276,480 @@ def patch_hic_summary_tables(connection, **kwargs):
...
+    action = ActionResult(connection, 'sync_users_oh_start')
+    my_auth = connection.ff_keys
+    sync_users_oh_check_result = action.get_associated_check_result(kwargs).get('full_output', {})
+    actions = sync_users_oh_check_result['actions']
+    user_list = get_oh_google_sheet()
+    add users
+    if actions.get('add_user'):
```
aschroed, line 2616: Specific comments would be useful here to distinguish adding removing info from User items from updating the spreadsheet.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2270,3 +2276,480 @@ def patch_hic_summary_tables(connection, **kwargs):
...
+        for a_user in actions['add_user']:
+            del a_user['lab_score']
+            del a_user['OH_lab']
+            a_user['viewing_groups'] = ["4DN"]
+            ff_utils.post_metadata(a_user, 'user', my_auth)
+    # delete users
```
aschroed, line 2622: Again better to say remove permission from user or something rather than delete.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2270,3 +2276,480 @@ def patch_hic_summary_tables(connection, **kwargs):
...
+            for c, key in enumerate(line):
+                col = c + 1
+                gs_write.append(gspread.models.Cell(row, col, key))
+        row = r + 2
+        # write values
+        for c, key in enumerate(line):
+            col = c + 1
+            gs_write.append(gspread.models.Cell(row, col, line[key]))
+    # #Write the cells to the worksheet
+    worksheet.update_cells(gs_write)
```
aschroed, line 2681: Is the worksheet automatically updated or does a copy need to be sent to OH?

aschroed (changes requested):

aschroed:
In general this looks good but we should discuss a final time.  Most of my comments are non-blocking and have more to do with updating the comments and renaming some things to be clearer but we should consider if it makes more sense to consider 4DN membership being the presence of the user.viewing_groups=4DN rather than trying to get that info from lab and awards.

aschroed (approved):

chalicelib/checks/wrangler_checks.py
```diff
@@ -2270,3 +2276,556 @@ def patch_hic_summary_tables(connection, **kwargs):
...
+            del a_user['lab_score']
+            if a_user.get('OH_lab'):
+                del a_user['OH_lab']
+            ff_utils.patch_metadata(a_user, user_uuid, my_auth)
+
+    # remove user's permissions from the data portal
+    if actions.get('delete_user'):
+        for a_user in actions['delete_user']:
+            # Delete the user permissions: submits_for, groups, viewing_groups and lab.
+            ff_utils.patch_metadata({}, a_user, my_auth, add_on='delete_fields=submits_for')
```
aschroed, line 2702: These 4 patch operations can be merged into a single by using the ff_utils.delete_field function eg.
`ff_utils.delete_field(user_uuid, 'submits_for, lab, viewing_groups, groups', key=my_auth`)

aschroed (approved):
Assuming your testing has worked as expected this looks good to me - one suggestion in the action regarding deleting all the credential fields in one go.

PR #389, opened 2021-01-27
modify badge check for single bio rep

andreacosolo:
This check adds a badge to ExpSets with a single biological replicate. Added an exception to account for false positives.

An ExpSet might have many replicates, but only one present in the database. This happens especially with imaging datasets, where only one replicate with raw data is show as an example, but processed results are derived from many replicates.
Such ExpSets are labeled with a specific [static header](https://data.4dnucleome.org/static-sections/85520ecd-8df0-4d29-bd78-d8efa1f12ecf/). ExpSets with this header to not receive the 'replicate-numbers' badge.

### DIFF
--- a/chalicelib/checks/badge_checks.py
+++ b/chalicelib/checks/badge_checks.py
@@ -363,9 +363,8 @@ def repsets_have_bio_reps(connection, **kwargs):
 
             # check if single biological replicate
             if len(rep_dict.keys()) == 1:
-                many_replicates = '/static-sections/85520ecd-8df0-4d29-bd78-d8efa1f12ecf/'
-                # this header labels an ExpSet with many replicates, but only one present in the database (typically imaging datasets)
-                if many_replicates in result.get('static_headers', []):  # skip false positive
+                # this tag labels an ExpSet with many replicates, but only one present in the database (typically imaging datasets)
+                if 'many_replicates' in result.get('tags', []):  # skip false positive
                     continue
                 audits[audit_key]['single_biorep'].append(result['@id'])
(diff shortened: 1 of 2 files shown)


chalicelib/checks/badge_checks.py
```diff
@@ -361,7 +361,12 @@ def repsets_have_bio_reps(connection, **kwargs):
             else:
                 audit_key = RELEASED_KEY
 
+            # check if single biological replicate
             if len(rep_dict.keys()) == 1:
+                many_replicates = '/static-sections/85520ecd-8df0-4d29-bd78-d8efa1f12ecf/'
```
aschroed, line 366: I wonder if it might be more easily maintained if you were to add a tag for this rather than relying on the static section.  Gives more flexibility to swap out to a similar but different static section.  It may be easier to remember to add the tag during submission and then eventually if this starts to appear more frequently set up a static header check to add the header based on a tag.  This is not blocking the merge and may not even be a good suggestion - just a thought I had.

aschroed (approved):
Looks good.  Left one comment for your consideration

chalicelib/checks/badge_checks.py
sbreiff: I  also like the idea of using a tag but am ok either way.

sbreiff (approved):

chalicelib/checks/badge_checks.py
andreacosolo: That's a good idea, definitely more flexible.

PR #399, opened 2021-02-16
removed dependency from expset_opf_unique_files_in_experiments

andreacosolo:


aschroed (approved):

andreacosolo:
Actually the dependency made sense, so I reverted the commit.
This PR only has minor edits to the documentation.

PR #416, opened 2021-03-29
update higlass registration check/action to utilize open_data_url if present

aschroed:
Because open_data_url is a calcprop that checks to see if a file is present in the open data bucket if the property is present we don't double check but take it's word for it.  Is that sufficient?  or should an attempt be made to call does_key_exist function by parsing said url?

Registration process updated to use open_data_url as filepath payload.


willronchetti (approved):
LGTM - should double check that higlass server will accept a full S3 URL (instead of just bucket/key)

alexander-veit:
Just for the record: This is not ready to be merged and will break the Higlass registration. I will upload a new version after I could do more extensive testing.

alexander-veit:
I think it should work now, but we have to test it again, when the missing extra files have been copied to the open data bucket.

PR #417, opened 2021-03-29
Add pi lab to missing labs

Lmercadom:


chalicelib/checks/wrangler_checks.py
```diff
@@ -2412,19 +2412,29 @@ def compare_record(existing_record, user, all_labs, all_grants, new=False):
...
             s = fuzz.token_sort_ratio(record['OH Lab'], disp.split(',')[0])
             if s > score:
                 best = disp
                 score = s
         if score > 73:
             lab = [i['@id'] for i in all_labs if i['display_title'] == best][0]
-        return lab, score
+
+        if not lab:
+            oh_grant = record.get('OH Grant', '')
```
aschroed, line 2430: So here you are getting a value from the OH spreadsheet, right?  And if there isn't one then oh_grant is the empty string.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2412,19 +2412,29 @@ def compare_record(existing_record, user, all_labs, all_grants, new=False):
...
+            grant = [i for i in all_grants if i['name'].endswith(oh_grant)]
```
aschroed, line 2431: So here is there was not a value in the OH Grant field then you will end up with a list of all the grants because you are looking for any award that ends with the empty string which is all of them.
i.e. if oh_grant = '' then if i['name'].endswith(oh_grant) will be true.
Is that what you intend? Because you just grab the first one for assigning the lab.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2683,6 +2695,8 @@ def sync_users_oh_start(connection, **kwargs):
             del a_user['lab_score']
             if a_user.get('OH_lab'):
                 del a_user['OH_lab']
+            if a_user.get('log'):
```
aschroed, line 2698: It probably doesn't happen or you would have already seen this issue but what happens if a_user['log'] is None or a falsy value - it wont be deleted from a_user.  I think likely the better way to do this kind of delete when removing items that are not real properties so leaving them in will cause the post or patch to fail is to test if the key exists at all and if it does delete it:
eg. 
`if 'OH_lab' in a_user:
      del a_user['OH_lab']
`
That way you never have a problem if a key exists but the value doesn't.

aschroed (changes requested):
I left one comment where it is a bit confusing if the check is always doing the expected thing.  If you can address that then the other comment about testing for the key is while I think the better way to do those tests not a blocking comment.

aschroed (approved):

aschroed:
Looks good.

PR #419, opened 2021-04-01
remove experimental/contributing lab check for opf

andreacosolo:
This check was adding the experimental lab as `contributing_labs` to the opf if the file's lab is not the same as the experiment/set's lab.
Now there is a calcprop that adds the `experimental_lab` under track_and_facet_info, so this check is not needed anymore.


andreacosolo:
We eventually decided that it is correct to have the experimental lab be also contributing lab to the opf produced by someone else on their data. So we keep the check. Also, this way future files will be consistently attributed.

PR #420, opened 2021-04-05
ChipSeqPipeline

KorayKirli:


chalicelib/checks/wfr_encode_checks.py
```diff
@@ -0,0 +1,490 @@
...
+                        check.brief_output.append(set_summary)
+                        check.full_output['skipped'].append({set_acc: set_summary})
+                        continue
+                run_ids = {'run_name': set_acc,
+                           'desc': a_set.get('description', '')}
+                parameters = {
+                    "chip.pipeline_type": target_type,
+                    "chip.paired_end": chip_p,
+                    "chip.choose_ctl.always_use_pooled_ctl": True,
+                    "chip.qc_report.name": run_ids['run_name'],
```
Lmercadom, line 394: There is an issue with the display_title of some of the ChIP-seq Quality Metric objects when using the "chip.qc_report.name” parameter. The calculated property takes the value from the 'name' field created by this parameter and displays it. We saw that some runs did not make use of this parameter and it worked fine. They also did not make use of the parameter "chip.qc_report.desc”. So, do not think we need to make use of these parameters.

KorayKirli: How did that get the name and description if not from the run parameters? Can you share an example for the ones that worked fine without the parameters?

If there are not used at all, feel free to take them out.

PR #426, opened 2021-04-28
Fix user sync pending action

Lmercadom:


aschroed (approved):
Hi Luisa - looks good to me.  While catching errors and reporting them is always a good idea was there a case that precipitated you to add these?  In other words were there user posts failing and if so does it make sense as to why?

PR #431, opened 2021-05-14
Repair ES Domain

willronchetti:
- Fixes ES checks to use correct domain names

### DIFF
--- a/chalicelib/checks/system_checks.py
+++ b/chalicelib/checks/system_checks.py
@@ -24,6 +24,12 @@ from .helpers.confchecks import *
 
 
 # XXX: put into utils?
+FF_BLUE_ES_CLUSTER_DOMAIN = 'fourfront-blue-6-8'
+FF_GREEN_ES_CLUSTER_DOMAIN = 'fourfront-green-6-8'
+PROD_ES_CLUSTERS = [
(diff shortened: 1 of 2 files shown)


chalicelib/checks/system_checks.py
```diff
@@ -63,6 +69,14 @@ def elastic_search_space(connection, **kwargs):
     return check
 
 
+def resolve_es_domain_name(connection):
+    """ Small helper that acquires health page on the given ff_env and resolves blue vs. green. """
+    if 'blue' in connection.ff_es:
```
aschroed, line 74: Likely no issue but at all safer to check for 'fourfront-blue' - or is the idea if cgap blue green ever gets implemented this allows easier re-use?

aschroed (approved):
Looks good.  Have not tested.

chalicelib/checks/system_checks.py
willronchetti: This is ok but I notice now my docstring is out of date - I originally implemented this going to health page, but realized foursight has the ES URL as well, so it's unneeded. Will update that at least. This kind of utility could be used for CGAP blue/green but we probably wont be doing a blue/green setup for CGAP anyway.

PR #434, opened 2021-06-08
added arima nz to wfr check

SooLee:


aschroed (approved):
Bump the version before merging?

PR #437, opened 2021-06-16
change from using fuzzywuzzy to py_stringmatching

aschroed:


chalicelib/checks/wrangler_checks.py
```diff
@@ -1176,7 +1176,8 @@ def users_with_doppelganger(connection, **kwargs):
             cases.append(log)
         # if not, compare names
         else:
-            score = fuzz.token_sort_ratio(us1['display_title'], us2['display_title'])
+            matcher = stringmatch.Levenshtein()
+            score = matcher.get_sim_score(us1['display_title'], us2['display_title']) * 100
```
netsettler, line 1180: I assume the `* 100` means it's now a float 0.0-1.0. Was it previously a float at higher scale, or was it an integer? This will affect:
* whether 85.12345 if rounded to the nearest integer is greater than 85, because `score > 85` below will be true for `85.12345` but not for `85`.
* whether the part below where you use `'log': 'has similar names ({}/100)'.format(str(score))` and whether it shows a possibly very long decimal part.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2424,8 +2425,9 @@ def find_lab(record, all_labs, all_grants):
         score = 0
         best = ''
         log = []
+        matcher = stringmatch.Levenshtein()
         for disp in all_lab_names:
-            s = fuzz.token_sort_ratio(record['OH Lab'], disp.split(',')[0])
+            s = matcher.get_sim_score(record['OH Lab'], disp.split(',')[0]) * 100
```
netsettler, line 2430: Same potential issues here with `* 100`. Maybe or maybe not OK according to same details as above, I guess.

netsettler (commented):
I'm giving a provisional approval here. The technique seems sound, but I did have one concern that I'd like you to look at carefully before you close this to make sure you're getting semantic equivalence here (if it's in fact needed, as for some situations that might be overkill). I'll trust your judgment on this. I just wanted to flag the issue.

chalicelib/checks/wrangler_checks.py
aschroed: I added a round().  The actual cutoff is arbitrary so rounding should not significantly affect anything but will potentially tidy up output.

PR #442, opened 2021-06-28
Update the ontology check

aschroed:
- REVERTED THIS - I was wrong about this bit.  The re to parse the date out of the string needed to be updated to look for years beyond 2020 (depending on format of date in string)
- The svn URL from which UBERON version info was retrieved appears to no longer be updated.  Switched to get head of base UBERON ontology from purl URL.
- version bump

### DIFF
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -1401,7 +1401,7 @@ def check_for_ontology_updates(connection, **kwargs):
                 continue
             else:
                 # looks for date string in versionIRI line
-                match = re.search('(2[1-5])?([0-9]{2})-[0-9]{2}-(2[1-5])?[0-9]{2}', vline)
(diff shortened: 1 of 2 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -1368,42 +1368,50 @@ def check_for_ontology_updates(connection, **kwargs):
...
                 continue
-            if 'versionIRI' in owl.text:
-                idx = owl.text.index('versionIRI')
-                vline = owl.text[idx:idx+150]
-                if 'releases'in vline:
-                    vline = vline.split('/')
-                    v = vline[vline.index('releases')+1]
+            else:
+                # looks for date string in versionIRI line
+                match = re.search('(2[1-5])?([0-9]{2})-[0-9]{2}-(2[1-5])?[0-9]{2}', vline)
```
sbreiff, line 1404: This regex needs to be reverted to the previous one, I think. I can't successfully pick up dates such as '2021-06-05' with this.

sbreiff (approved):

PR #443, opened 2021-06-28
Remove fuzzywuzzy (again)

aschroed:
Functions that use builtin difflib SequenceMatcher look to be sufficient for these purposes - tested with the entire ignore list from doppleganger check .

### DIFF
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -1095,7 +1095,7 @@ def get_tokens_to_string(s):
 
 
 def string_label_similarity(string1, string2):
-    """ compares sorted tokens for similarity
+    """ compares concantenate token strings for similarity
         simple tokenization - return a score between
         0-1
     """
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -1085,19 +1085,22 @@ def finalize_user_pending_labs(connection, **kwargs):
     return action
 
 
-def get_and_sort_tokens_to_string(s):
-    tokens = [t.lower() for t in re.split(r'\W+', s) if t]
-    tokens.sort()
+def get_tokens_to_string(s):
+    """ divides a (potentially) multi-word string into tokens - splitting on whitespace or hyphens
+       (important for hyphenated names) and lower casing
+       returns a single joined string of tokens
+    """
+    tokens = [t.lower() for t in re.split(r'[\s-]', s) if t]
     return ''.join(tokens)
 
 
 def string_label_similarity(string1, string2):
     """ compares sorted tokens for similarity
-        simple tokenization - then sorting and return a score between
+        simple tokenization - return a score between
         0-1
     """
-    s1cmp = get_and_sort_tokens_to_string(string1)
-    s2cmp = get_and_sort_tokens_to_string(string2)
+    s1cmp = get_tokens_to_string(string1)
+    s2cmp = get_tokens_to_string(string2)
     return SequenceMatcher(None, s1cmp, s2cmp).ratio()
 
 
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -2453,8 +2453,7 @@ def sync_users_oh_status(connection, **kwargs):
         log = []
         # matcher = stringmatch.Levenshtein()
         for disp in all_lab_names:
-            # s = round(matcher.get_sim_score(record['OH Lab'], disp.split(',')[0]) * 100)
-            score = round(string_label_similarity(record['OH Lab'], disp.split(',')[0]) * 100)
+            s = round(string_label_similarity(record['OH Lab'], disp.split(',')[0]) * 100)
             if s > score:
                 best = disp
                 score = s
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -7,8 +7,7 @@ import datetime
 import time
 import itertools
 import random
-from fuzzywuzzy import fuzz
-# import py_stringmatching as stringmatch
+from difflib import SequenceMatcher
 import boto3
 from .helpers import wrangler_utils
 from collections import Counter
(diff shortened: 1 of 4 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -1086,6 +1085,22 @@ def finalize_user_pending_labs(connection, **kwargs):
     return action
 
 
+def get_and_sort_tokens_to_string(s):
+    tokens = [t.lower() for t in re.split(r'\W+', s) if t]
+    tokens.sort()
+    return ''.join(tokens)
+
+
+def string_label_similarity(string1, string2):
```
netsettler, line 1094: So this is telling me how similar two ordered series of strings are? Why are the tokens sorted? This seems quite well-described and it does what it does, but then when you use it below, it seems like it will do the wrong thing. Are you sure you want a function that _does_ sorting?

chalicelib/checks/wrangler_checks.py
```diff
@@ -1177,9 +1192,7 @@ def users_with_doppelganger(connection, **kwargs):
             cases.append(log)
         # if not, compare names
         else:
-            # matcher = stringmatch.Levenshtein()
-            # score = round(matcher.get_sim_score(us1['display_title'], us2['display_title']) * 100)
-            score = fuzz.token_sort_ratio(us1['display_title'], us2['display_title'])
+            score = round(string_label_similarity(us1['display_title'], us2['display_title']) * 100)
```
netsettler, line 1195:
```
>>> s1="""'Arthur Dean' <arthur.dean@bar.foo.com>"""
>>> s2="""'Dean Arthur' <dean.arthur@foo.bar.com>"""
>>> t1=[t.lower() for t in re.split(r'\W+', s1) if t]
>>> t2=[t.lower() for t in re.split(r'\W+', s2) if t]
>>> t1
['arthur', 'dean', 'arthur', 'dean', 'bar', 'foo', 'com']
>>> t2
['dean', 'arthur', 'dean', 'arthur', 'foo', 'bar', 'com']
>>> t1.sort()
>>> t2.sort()
>>> t1
['arthur', 'arthur', 'bar', 'com', 'dean', 'dean', 'foo']
>>> t2
['arthur', 'arthur', 'bar', 'com', 'dean', 'dean', 'foo']
```

Also:

```
>>> s1="""How to turn lead into gold"""
>>> s2="""How to turn gold into lead"""
>>> s3="""To turn into lead gold: how?"""
>>> t1=[t.lower() for t in re.split(r'\W+', s1) if t]
>>> t2=[t.lower() for t in re.split(r'\W+', s2) if t]
>>> t3=[t.lower() for t in re.split(r'\W+', s3) if t]
>>> t1
['how', 'to', 'turn', 'lead', 'into', 'gold']
>>> t2
['how', 'to', 'turn', 'gold', 'into', 'lead']
>>> t3
['to', 'turn', 'into', 'lead', 'gold', 'how']
>>> t1.sort()
>>> t2.sort()
>>> t3.sort()
>>> t1
['gold', 'how', 'into', 'lead', 'to', 'turn']
>>> t2
['gold', 'how', 'into', 'lead', 'to', 'turn']
>>> t3
['gold', 'how', 'into', 'lead', 'to', 'turn']
```
Are those really things you want to count as dopplegangers? It might be that you're just preserving some prior status quo where sorting was done, but I'm not sure. If not, I would just ask you to make sure you're accomplishing what you think you are meaning to accomplish here. I don't know the requirements so it's hard for me to evaluate. At some level, all code just does what it's programmed to do. :)

netsettler (approved):
Some unit testing would help this because at least in the absence of documentation it'd be clearer to me what you meant this to cover. 

I'm going to make a provisional approval here. It's possible this is doing what you want it to do, and you can go ahead. If it doesn't do what you want, you'll know better than I. 

f you decide you want the sort, please add a comment where the use of it is (which is more important than in the function) and explain why sorting is good there. If you don't want the sorting, I assume you'll change and rename the function. 

But I'm happy to take a look again if you make some other change with enough substance that you want a second pair of eyes. Let me know.

aschroed:
> Some unit testing would help this because at least in the absence of documentation it'd be clearer to me what you meant this to cover.
> 
I agree but we've never had unit testing framework set up for foursight checks - current tests are just for core app functionality.  This was a design decision by Carl when he first wrote the app but I think Soo may have  made changes to cgap-foursight and -core to facilitate testing?  If so I think it worth incorporating into fourfront-foursight as possible but outside the scope of this PR.

> I'm going to make a provisional approval here. It's possible this is doing what you want it to do, and you can go ahead. If it doesn't do what you want, you'll know better than I.

> 
> f you decide you want the sort, please add a comment where the use of it is (which is more important than in the function) and explain why sorting is good there. If you don't want the sorting, I assume you'll change and rename the function.
> 
Good point - I've removed the sort and by manually testing the most 'complex' current cases I think that is right call.  The current scoring cutoffs work as expected for the cases I tested but it is definitely worth keeping an eye on.

chalicelib/checks/wrangler_checks.py
```diff
@@ -2438,7 +2454,7 @@ def find_lab(record, all_labs, all_grants):
         # matcher = stringmatch.Levenshtein()
         for disp in all_lab_names:
             # s = round(matcher.get_sim_score(record['OH Lab'], disp.split(',')[0]) * 100)
-            s = fuzz.token_sort_ratio(record['OH Lab'], disp.split(',')[0])
+            score = round(string_label_similarity(record['OH Lab'], disp.split(',')[0]) * 100)
```
netsettler, line 2457: You changed "s =" to "score =". Is that intended? Looks like a bug.

netsettler:
On unit tests, I would not care if you just added a file that only tested that one utility function. These things are a work in progress. It doesn't have to be in this PR. But never let the fact that you could make more become a reason not to do whatever modest tests you might have time to do.

chalicelib/checks/wrangler_checks.py
aschroed: Great catch - thanks!

PR #445, opened 2021-07-20
Fix bam re motif key error and add Arima re motif

clarabakker:
Replaced the template overwrite step in step_settings method of wfrset_utils.py for cases where the dict does not yet have a key in the dict. E.g. if template = {'parameters': {}}, overwrite = {'parameters': {'motif': {'common_enz': 'MboI'}}}, then template['parameters']['common_enz'] does not exist to be overwritten.

Also added Arima motif for bam_re_check.

### DIFF
--- a/chalicelib/checks/helpers/wfrset_utils.py
+++ b/chalicelib/checks/helpers/wfrset_utils.py
@@ -450,5 +450,10 @@ def step_settings(step_name, my_organism, attribution, overwrite=None):
     if overwrite:
         for a_key in overwrite:
             for a_spec in overwrite[a_key]:
-                template[a_key][a_spec] = overwrite[a_key][a_spec]
+                # if the key value is a dictionary, set default and use update
+                if isinstance(overwrite[a_key][a_spec], dict):
+                    template[a_key].setdefault(a_spec, {}).update(overwrite[a_key][a_spec])
+                # if it is string array bool, set the value
+                else:
+                    template[a_key][a_spec] = overwrite[a_key][a_spec]
     return template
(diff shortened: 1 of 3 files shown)


chalicelib/checks/helpers/wfrset_utils.py
```diff
@@ -450,10 +450,5 @@ def step_settings(step_name, my_organism, attribution, overwrite=None):
     if overwrite:
         for a_key in overwrite:
             for a_spec in overwrite[a_key]:
-                # if the key value is a dictionary, use update
-                if isinstance(overwrite[a_key][a_spec], dict):
-                    template[a_key][a_spec].update(overwrite[a_key][a_spec])
-                # if it is string array bool, set the value
-                else:
-                    template[a_key][a_spec] = overwrite[a_key][a_spec]
+                template[a_key][a_spec] = overwrite[a_key][a_spec]
```
aschroed, line 453: Have you confirmed that there are never any existing values in the template[a_key][a_spec] dictionary in any pipeline so the update function is not needed.  This will replace anything that already exists and it is curious that this was used previously suggesting that there may have been a use case? 

If you are convinced that this is sufficiently tested and won't break other pipelines then everything else looks good to me.

And are we always

aschroed (approved):

chalicelib/checks/helpers/wfrset_utils.py
aschroed: Consider this conditional approval - I just want you to be sure that this change will work for all cases.

PR #449, opened 2021-08-13
Fix for ChIP-seq logging via stepper and for peak caller for some experiments (TF target w/o control)

clarabakker:
Without a specified organism, stepper fn defaults to human, which can result in confusing logs. To fix, organism added to stepper calls. Also, experiments with a transcription factor target and no control should use macs2 for peak calling to avoid errors. To fix, added a conditional to specify this parameter.


aschroed (approved):
Changes look fine to me but I am not familiar enough with the pipeline to know if there are any other considerations - for example, what if the target_type is not 'tf' and there is no control?  I guess that is OK but don't know enough to be sure so you should check with Soo.

PR #450, opened 2021-08-24
Minor tweak to change_in_item_count check

aschroed:
Minor tweak to change_in_item_count check so would not report fail when finding purged MicroscopeConfigurations; version bump


clarabakker (approved):
Looks good to me!

PR #456, opened 2021-09-20
Generalize protected data management

andreacosolo:
Initially only HeLa data was restricted. Now any human individual can have a protected_data flag. Foursight checks are updated accordingly.

### DIFF
--- a/chalicelib/check_setup.json
+++ b/chalicelib/check_setup.json
@@ -1552,8 +1552,8 @@
             }
         }
     },
-    "released_hela_files": {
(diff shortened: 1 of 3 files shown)


aschroed (approved):
Looks good.  I guess we want to make sure the fourfront change gets in and the Hela individual get's patched before this gets merged?  Or I guess it might be a good test?

andreacosolo:
[improve restricted files management](https://trello.com/c/xkCT0tM5/125-improve-restricted-files-management)

PR #462, opened 2021-11-05
removed workflow settings specifying EC2 micros

clarabakker:
Also set default minimum for merge-pairs and reformatted checks for readability


alexander-veit (approved):
Looks good to me.

PR #463, opened 2021-11-08
ignore url checks to biorxiv

andreacosolo:
The static section URL check tries to send a GET request to biorxiv, (either directly or through doi.org), which fails because biorxiv doesn’t like these automated requests anymore.
Skip requests to biorxiv and ignore 403s that come from doi, so that the check gets green again.

### DIFF
--- a/chalicelib/checks/audit_checks.py
+++ b/chalicelib/checks/audit_checks.py
@@ -437,8 +437,9 @@ def check_help_page_urls(connection, **kwargs):
             try:
                 request = requests.get(url.replace('&amp;', '&'), timeout=2)
                 if request.status_code == 403 and 'doi.org'.lower() in url:
-                    # requests to doi.org that get redirected to biorxiv will fail with 403 - skip
-                    continue
+                    # requests to doi.org that get redirected to biorxiv fail with 403
+                    addl_exceptions.setdefault(result['@id'], {})
+                    addl_exceptions[result['@id']][url] = str(403)
                 elif request.status_code not in [200, 412]:
                     broken_links.append((url, request.status_code))
                 elif request.status_code == 504:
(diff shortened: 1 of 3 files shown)


chalicelib/checks/audit_checks.py
```diff
@@ -425,6 +425,9 @@ def check_help_page_urls(connection, **kwargs):
         for url in urls:
             if url.startswith('mailto'):
                 continue
+            elif 'biorxiv.org'.lower() in url:
```
sbreiff, line 428: should this be `elif 'biorxiv.org' in url.lower():`? biorxiv.org is already lowercase.

chalicelib/checks/audit_checks.py
```diff
@@ -433,11 +436,12 @@ def check_help_page_urls(connection, **kwargs):
                 continue
             try:
                 request = requests.get(url.replace('&amp;', '&'), timeout=2)
-                if request.status_code not in [200, 412]:
+                if request.status_code == 403 and 'doi.org'.lower() in url:
```
sbreiff, line 439: here too doi.org is already lowercase, should this be `url.lower():`?

sbreiff (commented):
I commented on your use of `.lower()` in a couple lines. In addition, I see that the poetry.lock file has been updated - did you commit this on purpose? Usually we only update this when we need to update dependencies, but if the tests didn't pass until you committed this then it makes sense to update it.

chalicelib/checks/audit_checks.py
andreacosolo: That's what I meant, thanks for catching that!

andreacosolo:
> I see that the poetry.lock file has been updated - did you commit this on purpose? Usually we only update this when we need to update dependencies, but if the tests didn't pass until you committed this then it makes sense to update it.

When running the check locally, it didn't work until I did `poetry update`. Since I saw that tests didn't pass, I was hoping that this would fix them. But it looks like test_get_s3_utils is still failing. @willronchetti any idea why?

PR #465, opened 2021-11-12
Add CUT&RUN to wfr_checks

clarabakker:
Manage CUT&RUN runs via foursight. Includes associated workflow settings and adjustments to helper functions.


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "foursight"
-version = "1.6.8"
+version = "1.6.9"
```
dmichaels-harvard, line 3: This is very strange to me. This PR/branch is supposed to be from master but the master pyproject.tom has the version "4.7.1" - and other parts of this don't look right to me (e.g. packages/include should be "chalicelib_fourfront" and not "chalicelib") ... was this (cut_tib) branch really created from master?

dmichaels-harvard: Nevermind I realized this PR is very very old.

PR #466, opened 2021-11-16
update contributing lab check

andreacosolo:
- Improved the check to deal with a list of (instead of just one) experiments or experiment sets linked to an other_processed_file.
- Renamed some variables for improved clarity.

### DIFF
--- a/chalicelib/checks/wrangler_checks.py
+++ b/chalicelib/checks/wrangler_checks.py
@@ -1962,69 +1962,62 @@ def check_opf_lab_different_than_experiment(connection, **kwargs):
               '&track_and_facet_info.experiment_bucket%21=processed+file' +
               '&field=experiment_sets&field=experiments' +
               '&field=lab&field=contributing_labs' + from_date_query)
-    result = ff_utils.search_metadata(search, key=connection.ff_keys)
+    other_processed_files = ff_utils.search_metadata(search, key=connection.ff_keys)
 
-    opf = {'to_patch': [], 'problematic': []}
-    exp_set_uuids = []  # Exp or ExpSet uuid list
-    for res in result:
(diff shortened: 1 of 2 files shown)


chalicelib/checks/wrangler_checks.py
```diff
@@ -1962,69 +1962,62 @@ def check_opf_lab_different_than_experiment(connection, **kwargs):
...
+    for opf in other_processed_files:
+        if opf['@id'] in [opf_probl['@id'] for opf_probl in output_opfs['problematic']]:
+            # skip problematic files
+            continue
+        labs_to_add = []
+        opf_exp_set_labs = list(set([es_uuid_2_lab[exp_set] for exp_set in opf['exp_set_uuids']]))
+        if len(opf_exp_set_labs) == 1 and opf_exp_set_labs[0] == opf['lab']['uuid']:
+            # lab of exp/set is the same as lab of opf
+            continue
+        elif opf.get('contributing_labs'):
```
aschroed, line 1996: Should there be an else clause?  I'm not fully sure on what is happening here (and I can pull down the branch to check if needed) but it seems that either you only have one lab and if it's the same then we're good or you have 'contributing_labs'.  Can there be cases when len(opf_exp_set_labs) !=1 or opf_exp_set_labs[0] != opf['lab'] and there are not 'contributing_labs' where you want to do something?

aschroed (approved):
I added a comment just because I did not take the time to fully look at what was happening.  If you'd like me to explore further let me know.

chalicelib/checks/wrangler_checks.py
andreacosolo: You're right, this is a leftover from the previous code. It should be `else` instead.

PR #468, opened 2021-11-29
Upd hg chk bug

aschroed:
Update a higlass function that adds viewconfs to items - the code did not expect there to be existing static content that may lack the 'description' property so an exception was thrown when that case was encountered.  

The fix maintains any existing static content by replacing the object (or embedded) view of the content with its uuid but if an existing autogenerated higlass viewconf is found replaces the content with the newly generated viewconf.

### DIFF
--- a/chalicelib/checks/higlass_checks.py
+++ b/chalicelib/checks/higlass_checks.py
@@ -186,11 +186,13 @@ def add_viewconf_static_content_to_file(connection, item_uuid, higlass_item_uuid
     # Look through the static content to see if this section exists already.
     reuse_existing = False
     for sc in patched_static_content:
-        if sc["description"] == "auto_generated_higlass_view_config":
+        if "description" in sc and sc["description"] == "auto_generated_higlass_view_config":
             sc.update(new_sc_section)
             reuse_existing = True
-            break
-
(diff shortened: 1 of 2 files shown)


andreacosolo (approved):
This part of the code looks good to me, assuming you have tested it. I'm just not sure how it fits with the condition in L509-L512 (only replaces content with uuid if description is 'auto_generated_higlass_view_config'). Should that be more generic?
https://github.com/4dn-dcic/foursight/blob/567ccf1f9f0c2535c62e07d3965ef16f40d96810/chalicelib/checks/higlass_checks.py#L509-L512
Anyway, I see that this function is called in different places, so this looks like the best place to do the uuid replacement and it's probably not worth editing other parts of the check if it works already.

aschroed:
Interesting catch - this code is quite opaque.  I think this should be OK though as the function I modified is called from the function containing the lines that you pointed out so converting the embedded version to uuid is done in the function I modified - perhaps one day in the fullness of time someone can go back and refactor this code so it is clearer but hopefully my fix should catch most if not all things of this type.  I will go ahead and merge unless you think I've misunderstood the issue or have other concerns.  Thanks!

PR #469, opened 2022-01-13
Check md5sum on released files

andreacosolo:
We want to make sure that all released files have md5sum. Files without previous md5 runs can be run automatically, while those with pre-existing md5run only trigger a warning and should be resolved manually. The statuses checked include pre-release and other public ones.

### DIFF
--- a/chalicelib/checks/wfr_checks.py
+++ b/chalicelib/checks/wfr_checks.py
@@ -217,7 +217,7 @@ def md5run_released_files(connection, **kwargs):
         return check
     # Build the query
     statuses = ['pre-release', 'released',  # 'released to project', 'archived to project'
-                'archived', 'replaced']
+                'uploaded', 'archived', 'replaced']
     query = '/search/?type=File&md5sum=No+value' + ''.join(['&status=' + s for s in statuses])
 
     files = {}
--- a/chalicelib/checks/wfr_checks.py
+++ b/chalicelib/checks/wfr_checks.py
@@ -228,7 +228,7 @@ def md5run_released_files(connection, **kwargs):
 
     if files:
         check.status = 'WARN'
-        check.brief_output = {k: str(len(v)) + ' files' for k, v in files}
+        check.brief_output = {k: str(len(v)) + ' files' for k, v in files.items()}
         check.full_output = files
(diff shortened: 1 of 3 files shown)


chalicelib/checks/wfr_checks.py
```diff
@@ -204,6 +204,41 @@ def md5run_status(connection, **kwargs):
...
+    my_auth = connection.ff_keys
+    check.action = "md5run_start"
+    check.status = 'PASS'
+    # check indexing queue
+    check, skip = wfr_utils.check_indexing(check, connection)
+    if skip:
+        return check
+    # Build the query
+    statuses = ['pre-release', 'released',  # 'released to project', 'archived to project'
+                'archived', 'replaced']
```
aschroed, line 220: I know why you are not currently including '... to project' files.  What is the reason for not including uploaded files in the check?  When things are working correctly is not the status change done after the md5 has been run so uploaded files should always have an md5sum?  Just curious about the current state of things and if my understanding on what should happen is correct.

chalicelib/checks/wfr_checks.py
```diff
@@ -216,6 +251,7 @@ def md5run_start(connection, **kwargs):
     targets = []
     if kwargs.get('start_missing'):
         targets.extend(md5run_check_result.get('files_without_md5run', []))
+        targets.extend(md5run_check_result.get('released_without_md5run', []))
```
aschroed, line 254: I'm sure you already considered this but can you imagine any situation where we would want to be able to conditionally deal with these cases separately?  Probably not but just thought I'd throw it out there.

aschroed (approved):
Looks good - I made a couple of comments (I thought I had already left these but there were still marked as pending so maybe I never fully finished the review - but they are not blocking.

chalicelib/checks/wfr_checks.py
andreacosolo: You are right: for most files, the status is changed to uploaded only after md5sum is patched, so this would be already a good time to check if anything went wrong. ENCODE pipelines are an exception, since they won't have md5sum, but it would be good to check already when uploaded.
I just had to go back and patch also the uploaded files currently missing md5.

PR #470, opened 2022-01-25
Add ECS related checks

willronchetti:
- Adds/enables 3 ECS related checks on mastertest


chalicelib/checks/ecs_checks.py
```diff
@@ -0,0 +1,96 @@
...
+    check = CheckResult(connection, 'ecs_status')
+    full_output = {
+        'ECSMeta': {
+            'clusters': {}
+        }
+    }
+    client = ECSUtils()
+    cluster_arns = client.list_ecs_clusters()
+    for cluster_arn in cluster_arns:
+        if 'cgap' in cluster_arn:
```
netsettler, line 20: Is this check for `cgap` correct?

pyproject.toml
```diff
@@ -10,7 +10,7 @@ packages = [
 
 [tool.poetry.dependencies]
 python = ">=3.6.1,<3.7"
-dcicutils = "^3.1.0"
+dcicutils = "^3.6.1"
```
netsettler, line 13: Is this the right version? I guess you'll get a later version because of the caret, but I meant do you need a beta of dcicutils (for python 3.7/3.8 version of dcicutils, for example)? Maybe there's nothing in there you need.

netsettler (approved):
Minor comments on things to check. I don't _think_ I see anything blocking.

PR #472, opened 2022-01-26
Limit the number of runs

andreacosolo:
There is a rate limit on pulls from Docker Hub (currently 200 every 6 hours). While all workflows contribute to the total number of docker pulls, the easiest way to hit this limit is upon a large submission of fastq files. For each fastq file submitted, we run md5, fastQC and fastq-first-line workflows. Therefore, as a first attempt, we want to prevent having issues upon large submissions.

This PR adds a new function to check the number of workflow runs in the database created in the past 6 h (including deleted ones). For the three pipelines mentioned, only a reduced number of files is sent to the check output and action. Note that this happens silently: a message is printed only if no file can be run.
The check scheduling is also adjusted to that the three run at different times.

### DIFF
--- a/chalicelib/check_setup.json
+++ b/chalicelib/check_setup.json
@@ -872,7 +872,7 @@
         "title": "l) Fastq first line",
         "group": "Pipeline checks",
         "schedule": {
-            "hourly_checks_2": {
+            "hourly_checks_3": {
                 "data": {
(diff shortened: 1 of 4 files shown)


aschroed (approved):
I think this looks good.  I guess there is not really a good way to test this before deploying.

PR #473, opened 2022-02-01
Enzyme error fix

sbreiff:
The bam_re_status check had been erroring because of some files that were expected to have a restriction enzyme but did not, resulting in an error when the function attempted to perform a join() on a list containing None values. 

To fix this I added some separate logic for storing and outputting files without restriction enzymes, which also get skipped by the action.


aschroed (commented):
My only thought is that this maybe should still be a WARN status for the check and that we could add 'skip_processing' tags to the relevant files and query so the check doesn't run on them - that way we don't miss unexpected problems? If there is some reason that this is a bad idea then never mind, otherwise I think it should be fairly straightforward to modify in this way and I would be willing to look at it tomorrow if you don't have time.

aschroed (approved):

PR #474, opened 2022-02-02
Tweak repno badge chk

aschroed:
I modified the brief output of the badge check that looks at replicate number badges.
The number of replicates with badges that do not need to change is so large that you can't view either the brief or full output.  
I modified the check so that at least we can see the brief output without looking at the json by reporting the number of sets with a badge for which there is no change needed rather than the @ids.  The items that will get badges added or removed are still shown, allowing us to confirm that the action should be run.


sbreiff (approved):
This looks good but I think we could make the same change to full_output as well.

PR #477, opened 2022-02-11
Update biorxivs

andreacosolo:
BioRxiv stores different versions of a manuscript. We should make sure that we have the most recent one in the database, especially regarding title and list of authors, since these metadata are used by another check in foursight to search in Pubmed when the manuscript is published.

### DIFF
--- a/chalicelib/check_setup.json
+++ b/chalicelib/check_setup.json
@@ -106,6 +106,20 @@
             }
         }
     },
+    "biorxiv_version_update": {
+        "title": "biorxiv version update",
+        "group": "Metadata checks",
+        "schedule": {
+            "monthly_checks": {
+                "data": {
+                    "dependencies": [],
+                    "kwargs": {
+                        "primary": true
+                    }
+                }
+            }
+        }
+    },
     "biosource_cell_line_value": {
         "title": "Cell line biosources without cell_line metadata",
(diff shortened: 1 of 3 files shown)


andreacosolo:
The last commit introduces a comparison between the current version in biorxiv and in the 4DN database, and only triggers an update when a difference is found. This is to prevent indexing items when not necessary (including Experiments and ExpSets linked to the Publication, which in turn are detected by other foursight checks as recently modified - e.g. higlass checks, hic table update check, ...). However, doing so introduces a little code duplication from the Publication _update() in fourfront.
We can discuss what is best and revert this commit in case.

aschroed:
Have you explored the possibility of adding a biorxiv_version property to Publication schema and querying with biorxiv api to compare?  Seems a potentially more rigorous way to catch revisions though if the info we display hasn't really changed and the link to biorxiv still works and points to the latest probably not necessary to do the re-index/update?

Otherwise this seems a reasonable way to look for updates and not sure there is a good way to avoid the code duplication.

aschroed (approved):
See my comment but otherwise looks good.

andreacosolo:
Wait to merge until addition of `version` in Publication schema is in production https://github.com/4dn-dcic/fourfront/pull/1602

PR #480, opened 2022-02-24
Register BED files of type FileReference to Higlass

alexander-veit:
BED files of type FileReference that have an `extra_file` beddb are now also registered to Higlass. The beddb is assumed to live in the wfoutput bucket.


andreacosolo (approved):
Looks good

PR #481, opened 2022-02-24
Replaced biosource.individual.organism with biosource.organism 

sbreiff:
Replaced biosource.individual.organism with biosource.organism wherever found due to changes in fourfront embeds


aschroed (approved):
Thanks for this!

clarabakker (approved):

PR #485, opened 2022-03-03
update RNAseq strandedness tags check

andreacosolo:
- expect `skip_processing` tag in the Experiment Set, rather than in the Experiment
- for improved performance, rewritten the check to do a single search for Files instead of recursively getting each single one


aschroed (approved):
Looks good!

PR #486, opened 2022-03-04
Fix higlass check

andreacosolo:


alexander-veit (approved):
Good catch!. Thanks for looking at it again.

PR #495, opened 2022-05-04
Fix user act gspread

aschroed:
Changed the call to gspread.Cell to work with latest version.


clarabakker (approved):
Looks good! Webdev check failure is expected for the time.

PR #499, opened 2022-06-21
Changes to use configurable env_utils

netsettler:


netsettler:
I did not fix the tests, so it's unsurprising the tests fail. They didn't work before either. But hopefully you can use `chalice local` and poke at it to see if it does what you want, @willronchetti.

willronchetti (approved):

PR #500, opened 2022-08-08
Run beddb on reference files

andreacosolo:


aschroed (approved):
Without having looked at the related code I guess this puts all the beddb files in the correct (wfoutput) bucket regardless of file_type?

andreacosolo:
That's right, that part is not changed. Therefore also beddb files deriving from FileReference end up in the wfoutput bucket. Tibanna was already modified to deal with this edge case where the `bed` FileReference is user-submitted (`files` bucket) and the deriving `beddb` `extra_files` is in `wfoutput` bucket.

PR #501, opened 2022-08-08
Updates for C4-826 to IDENTITY-ize Foursight.

dmichaels-harvard:


app.py
```diff
@@ -329,6 +329,42 @@ def delete_environment(environ):
...
 
 
+# dmichaels/2022-08-08:
+# For testing/debugging/troubleshooting.
+@app.route('/info/{environ}', methods=['GET'])
+def get_view_info_route(environ):
+    req_dict = app.current_request.to_dict()
+    domain, context = app_utils_obj.get_domain_and_context(req_dict)
+    is_admin = app_utils_obj.check_authorization(req_dict, environ)
+    return app_utils_obj.view_info(request=app.current_request, environ=environ, is_admin=is_admin, domain=domain, context=context)
```
willronchetti, line 339: Plan still to move these among others back to foursight-core?

willronchetti (approved):

app.py
dmichaels-harvard: Yes, but not now, I was playing with that in my `react-experiment` branch. And that (moving them) will also take care of that CHALICE_LOCAL question you had in the foursight-cgap PR-83.

PR #505, opened 2022-09-30
add static content from files in opf to the ignore list

andreacosolo:
This includes higlass viewconfig items that are generated for each file (not to confuse with those for the opf collection, that were already included).

Also, refactored the code for clarity and easier maintenance.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,9 +19,12 @@ Change Log
 =====
 * Added this CHANGELOG.rst.
 * Spruced up Foursight UI a bit (virtually all in foursight-core but mentioning here).
+
   * New header/footer.
+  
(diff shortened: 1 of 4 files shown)


andreacosolo:
[check_status_mismatch should look at higlass items in relation to the file status](https://trello.com/c/ptTHGk0C/158-checkstatusmismatch-should-look-at-higlass-items-in-relation-to-the-file-status)

andreacosolo:
Thanks Andy, I initially just preserved the previous functionality and did not think at this particular case. I agree that it's an unlikely scenario, but it was easy to add support for qc_lists too.
I've only tested the function so far, but have not yet run the entire foursight check. I still want to do that before merging.

PR #507, opened 2022-11-02
Check schedule edits

andreacosolo:
It seems unnecessary to run metadata-related checks on `staging`, as it shares the database with `data`. These include:
- static header checks
- badge checks
- audit checks

Additionally, some metadata-related checks that were scheduled to run on `all` environments are only needed for `data`, and not for non-production environments. This includes static headers, badges, unique titles in opf, missing publication, missing ontology term for cells, assay names, etc.

In general, it might be good to have checks available for various environments, but running them on a schedule doesn't seem always necessary. Therefore, in these cases where I removed `all`, I added `display: all`.

This PR also disables the staging deployment check, currently unused, by setting it to the manual_checks schedule.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,13 @@ foursight
 Change Log
 ----------
 
+2.1.2
+=====
+
+* Update check schedule to reduce the number of metadata-related checks running on
+  staging and non-production environments.
+
+
 2.1.1
 =====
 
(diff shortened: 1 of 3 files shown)


andreacosolo:
This does not remove completely all checks from `staging`, but only most of the metadata-related ones. 

Checks that remain for staging are:
- elasticsearch_s3_count_diff
- scale_down_elasticsearch_production
- scale_up_elasticsearch_production

I'll check with @willronchetti anyway.

willronchetti (approved):

PR #509, opened 2022-11-08
updated script wfr_utils.py for Hi-Chip

RahiNav:
Script wfr_utils.py modified:
Hi-ChIP': ["HiC_Pipeline_0.2.6", "HiC_Pipeline_0.2.7"]


chalicelib/checks/helpers/wfr_utils.py
```diff
@@ -206,7 +206,7 @@
     'sci-Hi-C': [''],
     'MC-3C': [''],
     'MC-Hi-C': [''],
-    'Hi-ChIP': [''],
+    'Hi-ChIP': ["HiC_Pipeline_0.2.6", "HiC_Pipeline_0.2.7"],
```
aschroed, line 209: There is no need to have version 0.2.6 listed - you should remove that version from the list.  Should also add the same comment as there is on Plac-seq i.e. # Preliminary - Released to network # NO-NORM.

aschroed (changes requested):
It looks like David just merged in his react updates so you should pull master and merge master into your branch.  

And then it looks like you actually need to add a specific check function and action for the experiment type.  You can just duplicate and rename in all the relevant places the check and action for plac-seq starts here: https://github.com/4dn-dcic/foursight/blob/e68e80aa5593eef63ec7dcd340a7d0a116a151a3/chalicelib_fourfront/checks/wfr_checks.py#L1202 and ends here: https://github.com/4dn-dcic/foursight/blob/e68e80aa5593eef63ec7dcd340a7d0a116a151a3/chalicelib_fourfront/checks/wfr_checks.py#L1255.  

You will also need to add the check to check_setup.json - again duplicate and change names for plac-seq https://github.com/4dn-dcic/foursight/blob/e68e80aa5593eef63ec7dcd340a7d0a116a151a3/chalicelib_fourfront/check_setup.json#L1279

And finally you need to bump the version in pyproject.toml and update the readme.

chalicelib_fourfront/checks/helpers/wfr_utils.py
```diff
@@ -206,7 +206,8 @@
     'sci-Hi-C': [''],
     'MC-3C': [''],
     'MC-Hi-C': [''],
-    'Hi-ChIP': [''],
+    # Preliminary - Released to network  # NO-NORM
```
andreacosolo, line 209: I'm not sure about the "Released to network". This needs to be reviewed, and maybe you might need to edit the comments also on the other targeted-Hi-C types.

Additionally, you might move this block right below plac-seq so that they are all together.

andreacosolo (commented):
Looks good, just a few minor edits suggested.

chalicelib_fourfront/checks/helpers/wfr_utils.py
aschroed: Agreed on moving the location so all the similar types are grouped together.  I think the 'released to network' comments are not quite accurate.  You might check with @clarabakker to see if these should be removed or updated.  They probably correspond to the various parameters in the checks and/or actions and reflect things like whether the generated files are automatically moved to the processed file tab or if they are kept in supplementary files and others like normalization or if there is an enzyme file used.  In any case Hi-ChIP should be treated exactly as Plac-seq so as long as you used that check/action as the template and didn't change any parameters might as well keep the same comment as for that pipeline and then do a wholesale review of the comments in a separate PR after getting Clara's input.

RahiNav: Ok, moving the Hi-ChIP block below plac-seq, keeping the comment as it is for now. I will check with Clara to update in another PR.

aschroed (commented):
I'd bump the minor version instead of patch version - i.e. 2.3.0 as you are adding a new feature - that is a new check and action.

CHANGELOG.rst
```diff
@@ -6,6 +6,10 @@ foursight
 ----------
 Change Log
 ----------
+2.2.1
+=====
+
+* Update to wfr_utils.py, wfr_checks.py and wfr_checks.py to add HiChIP pipeline information
```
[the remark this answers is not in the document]
andreacosolo: My previous comment was suggesting a replacement of line 12, not just an addition.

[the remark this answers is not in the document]
andreacosolo: My previous comment was suggesting a replacement of line 12, not just an addition.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "foursight"
-version = "2.2.0"
+version = "2.2.1"
```
andreacosolo, line 3:
```suggestion
version = "2.3.0"
```

andreacosolo (approved):

chalicelib_fourfront/checks/helpers/wfr_utils.py
clarabakker: Yes, the specifications in the comment should be determined by the parameters used in the hichip_status and hichip_start functions. Parameters for omitting restriction enzymes and normalization are in the call to [wfr_utils.check_hic()](https://github.com/4dn-dcic/foursight/blob/a9737bfc838790fd86af38c0985bf12b074f31d5/chalicelib_fourfront/checks/wfr_checks.py#L1290), so NO-NORM is correct here.

"Preliminary - Released to network" should be used depending on the `move_to_pc` parameter of [wfr_utils.start_tasks](https://github.com/4dn-dcic/foursight/blob/a9737bfc838790fd86af38c0985bf12b074f31d5/chalicelib_fourfront/checks/wfr_checks.py#L1309). Since `move_to_pc` is true, the processed files will not be preliminary, so I think this "Preliminary - Released to network" can be removed (unless we have any reason we would want to restrict the files). PLAC-seq is the same (not preliminary).

aschroed (approved):

clarabakker (changes requested):
Two small things! Has this been run on an EC2?

chalicelib_fourfront/checks/wfr_checks.py
```diff
@@ -1254,6 +1254,61 @@ def plac_seq_start(connection, **kwargs):
...
+    check = CheckResult(connection, 'hichip_status')
+    my_auth = connection.ff_keys
+    check.action = "hichip_start"
+    check.description = "run missing steps and add processing results to processed files, match set status"
+    check.brief_output = []
+    check.summary = ""
+    check.full_output = {'skipped': [], 'running_runs': [], 'needs_runs': [],
+                         'completed_runs': [], 'problematic_runs': []}
+    check.status = 'PASS'
+    exp_type = 'Hi-ChIP'
```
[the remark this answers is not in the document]
RahiNav: Thanks for catching this @clarabakker.
Also, working on testing this via EC2.

clarabakker (approved):

PR #510, opened 2022-11-11
Edit consistent replicate info check

andreacosolo:


andreacosolo:
[improve Inconsistent Replicate Info message (especially protocols)](https://trello.com/c/R5Caysuv/120-improve-inconsistent-replicate-info-message-especially-protocols)

aschroed (approved):
Looks good.  The only question I had was about when badges will be noticed to need updates.  If I understand your comment is it possible that in some cases if an item already has a badge, and the only thing that has changed is the value of the property and not the properties themselves then the badge won't be updated? That surely can't be the case as then how would the existing badges get updated?  I'm definitely missing something in your explanation and it wasn't entirely clear to me from the code changes but if you have tested to your satisfaction it looks good to me.

andreacosolo:
> in some cases if an item already has a badge, and the only thing that has changed is the value of the property and not the properties themselves then the badge won't be updated? That surely can't be the case as then how would the existing badges get updated? 

@aschroed You are right: if the only thing that changes is the value (and not the property), this check will not detect the change with default parameters. I initially thought that in such case the wrangler could use the ignore_details arg to fix this. The main reason for that was because I wanted to reduce "needless" get_metadata calls. However, I verified that this only adds a few seconds to the check run time, so probably it's fine to revert this change and always check messages entirely.

aschroed:
Ah I had missed the significance of the ignore_details arg - I leave it up to you if you want to keep as is or maybe just set to False initially and then if it does add significant overhead we could switch to ignore details by default as you have it now.

PR #512, opened 2022-11-16
Add DNase Hi-C to types that insulation and compartment pipelines run on

aschroed:
Added DNase-Hi-C to ins and comp pipes
set pseudo re_nz_size for DNaseI to 4 in insulation status chk to correctly set bin size


clarabakker (approved):
Looks good!

PR #513, opened 2022-11-18
Comments edited for wfr_utlis.py

RahiNav:
Edited comments for accepted_versions. 
Comments edited based on wfr_checks.py (for #NO-RE and #NO-NORM) and status of released experiment-sets associated with experiment type (for Official or Preliminary)


RahiNav:
Also for experiment type  like "single cell Repli-seq" and "DNA SPRITE" which are official but don't have any pipe mentioned:
'single cell Repli-seq': [''],
Do I need to add comments for these experiments as well?

PR #514, opened 2022-11-21
Errored runs chk edit

andreacosolo:
* Edited the check for errored workflow runs to only report recent ones.
  * By default in the past 30 days. This can be modified using the ``days_back`` arg.
  * Use any string different than a number to search all errored runs.


clarabakker (approved):
Works for me! Approved with version update, of course.

andreacosolo:
[errored runs is always yellow](https://trello.com/c/v5ceGz01/167-errored-runs-is-always-yellow)

PR #516, opened 2022-12-06
Support for check actions

dmichaels-harvard:


dmichaels-harvard:
Changes here are mostly the addition of an action argument to the check_function decorator for the checks, so that we can easily determine (for the UI) which checks have associated actions.

willronchetti (approved):

PR #522, opened 2023-01-25
limit beddb reruns for reference files

andreacosolo:
Pipelines normally do not start automatically in case (2 or more) previous runs are detected. This mechanism does not work for FileReference items, because they do not have workflow run inputs. Since bed2beddb pipeline can run also on FileReference items, this PR introduces a different way to check for pre-existing runs of the same pipeline on the same file.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ foursight
 Change Log
 ----------
 
+3.3.1
+=====
+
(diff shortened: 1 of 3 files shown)


andreacosolo:
[Track failed workflow runs on Reference Files](https://trello.com/c/VIhnFSn8/145-track-failed-workflow-runs-on-reference-files)

aschroed (approved):
Looks good to me.

PR #523, opened 2023-02-02
Update to foursight-core 3.3.1 and dcicutils 6.7.0.

dmichaels-harvard:


netsettler:
I see you merged this without review. I think the version of dcicutils you have might be too low to accommodate the fix to get is_c4_arn in, so you might have to re-issue this to get foursight-core to line up.

PR #528, opened 2023-05-02
ChIP-seq update to 2.1.6

clarabakker:
Modify wfr_encode_checks to run the updated (v2.1.6) ChIP-seq pipeline


clarabakker:
(Apologies for the wonky Ubuntu commits)

chalicelib_fourfront/checks/helpers/wfrset_utils.py
```diff
@@ -195,9 +195,9 @@ def step_settings(step_name, my_organism, attribution, overwrite=None):
     },
     {
         "app_name": "encode-chipseq-aln-chip",
-        "workflow_uuid": "4dn-dcic-lab:wf-encode-chipseq-aln-chip",
+        "workflow_uuid": "212a9c91-25d6-473f-b56b-8dd93958c580",
```
aschroed, line 198: despite the property name, might it be better to use the alias as was formerly done in case the workflow is added to a new environment (with a different uuid being generated) - perhaps unlikely but if at all possible?

chalicelib_fourfront/checks/helpers/wfrset_utils.py
```diff
@@ -213,9 +213,9 @@ def step_settings(step_name, my_organism, attribution, overwrite=None):
     },
     {
         "app_name": "encode-chipseq-aln-ctl",
-        "workflow_uuid": "4dn-dcic-lab:wf-encode-chipseq-aln-ctl",
+        "workflow_uuid": "4eb427f1-a7d5-4d74-8cfa-4c77f42d5b43",
```
aschroed, line 216: Same as previous comment

chalicelib_fourfront/checks/helpers/wfrset_utils.py
```diff
@@ -226,9 +226,9 @@ def step_settings(step_name, my_organism, attribution, overwrite=None):
     },
     {
         "app_name": "encode-chipseq-postaln",
-        "workflow_uuid": "4dn-dcic-lab:wf-encode-chipseq-postaln",
+        "workflow_uuid": "291d4c64-75de-434a-9d98-01f40d19e15e",
```
aschroed, line 229: see above

chalicelib_fourfront/checks/wfr_encode_checks.py
```diff
@@ -183,39 +195,46 @@ def chipseq_status(connection, **kwargs):
...
+                input_files['chip.bowtie2_idx_tar'] = '/files-reference/4DNFIMQPTYDY/'
                 input_files['chip.blacklist'] = '/files-reference/4DNFIZ1TGJZR/'
                 input_files['chip.chrsz'] = '/files-reference/4DNFIZJB62D1/'
-                input_files['additional_file_parameters'] = {"chip.bwa_idx_tar": {"rename": "GRCh38_no_alt_analysis_set_GCA_000001405.15.fasta.tar"}}
+                input_files['chip.ref_fa'] = '/files-reference/4DNFI823L888/'
+                input_files['additional_file_parameters'] = {"chip.bwa_idx_tar": {"rename": "GRCh38_no_alt_analysis_set_GCA_000001405.15.fasta.tar"}, "chip.bowtie2_idx_tar": {"rename": "GRCh38_no_alt_analysis_set_GCA_000001405.15.bowtie2Index.tar"}}
             if organism == 'mouse':
                 org = 'mm'
                 input_files['chip.bwa_idx_tar'] = '/files-reference/4DNFIZ2PWCC2/'
+                input_files['chip.bowtie2_idx_tar'] = '63e22058-79c6-4e24-8231-ca4afac29dda'
```
aschroed, line 206: Again using uuid here could be prone to problems on different environments?

chalicelib_fourfront/checks/wfr_encode_checks.py
```diff
@@ -158,15 +169,16 @@ def chipseq_status(connection, **kwargs):
                 for merge_case in input_list:
                     merge_enum += 1
```
aschroed, line 170: It may not matter but it could be confusing if this is incremented even though there is nothing in 'merge_case' - may want to move this after the if?

chalicelib_fourfront/checks/wfr_encode_checks.py
```diff
@@ -348,10 +357,15 @@ def chipseq_status(connection, **kwargs):
...
                     s2_input_files['chip.chrsz'] = '/files-reference/4DNFIZJB62D1/'
+                    s2_input_files['chip.ref_fa'] = '/files-reference/4DNFI823L888/'
+                    s2_input_files['chip.bowtie2_idx_tar'] = '/files-reference/4DNFIMQPTYDY/'
+
                 if organism == 'mouse':
                     org = 'mm'
                     s2_input_files['chip.blacklist'] = '/files-reference/4DNFIZ3FBPK8/'
                     s2_input_files['chip.chrsz'] = '/files-reference/4DNFIBP173GC/'
+                    s2_input_files['chip.ref_fa'] = '/files-reference/4DNFIC1NWMVJ/'
+                    s2_input_files['chip.bowtie2_idx_tar'] = '63e22058-79c6-4e24-8231-ca4afac29dda'
```
aschroed, line 368: again about using the uuid?

aschroed (approved):
I've left a few non-blocking comments - there may be good reasons for doing as you have done and if the comments don't make sense feel free to ignore.

Dealing with the pairing situations are complex and given that you have done extensive testing I expect that if there is a list index used that is not present thereby causing an exception then that is indicating a definite problem that should cause the check to error - looking at spots like Line232 and 263 in wfr_encode_checks.py.  Just wondering if a try/except to skip and report problematic cases might be warranted rather than having the whole check error?

chalicelib_fourfront/checks/helpers/wfrset_utils.py
clarabakker: Yes, I was aiming to make it consistent with other workflows, since I made the new UUIDs are consistent (for the current envs). The aliases also are the same for both new and old versions, which maybe also was a factor? It needs to know the new and old version are the same workflow, but only run the new one.

chalicelib_fourfront/checks/wfr_encode_checks.py
clarabakker: Oh wait, this was also intentional (to resolve problems between envs). I will test to see if it's still an issue

chalicelib_fourfront/checks/helpers/wfrset_utils.py
clarabakker: Oops, I see your confusion--the aliases are definitely not the same and the app name keeps the versions together. I do think it is more consistent with other workflows to use the uuid. It is also hard to query aliases for someone going from this workflow setting to the workflow on the portal, right?

PR #529, opened 2023-05-04
new check to automate patching of FOF-CT static section for chromatin datasets

RahiNav:
As we don't have a combined "chromatin tracing" experiment type, I added the different experiment sets (Multiplexed FISH, RNA FISH, SPT, OptoDroplet, DNA FISH, Immunofluorescence, Electron Tomography) under the same search. Let me know if this works or I can create the same check for different experiment types separately.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,10 @@ foursight
 Change Log
 ----------
 
+3.4.3
+=====
+* Added a new check for automate patching of FOF-CT static section for chromatin tracing datasets
+
 3.4.2
 =====
 * Version changes related to foursight-core changes for SSL certificate and Portal access key checking.
--- a/chalicelib_fourfront/check_setup.json
+++ b/chalicelib_fourfront/check_setup.json
@@ -1500,6 +1500,26 @@
             }
         }
     },
+    "prepare_static_headers_Chromatin_Tracing": {
+        "title": "Chromatin Tracing",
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,10 @@ foursight
 Change Log
 ----------
 
+3.4.3
+=====
+* Added a new check for automate patching of FOF-CT static section for chromatin tracing datasets
+
```
aschroed, line 13: It's good to mention which file the check is in.

aschroed (changes requested):
The only type of experiment that currently routinely should get an FOF header is 'multiplex_FISH', right?  A recent RNA-FISH dataset used a variation of the FOF format but we included comments there to say that it did not exactly comply so I think they don't necessarily need the header in that case.  
Basically both searches should be modified so they are only looking for 'multiplexed FISH'.  If we want to add the headers to other experiments we can do that by hand.

BTW - it looks like all these header checks are structured the same way but most don't need the 'remove' search and consequently I think they are doing more work than necessary.  Changing the type of experiment is rare (and very difficult/involved) so not many of these sections will ever be automatically removed and so that search is done for no reason.  It may be worth re-factoring these checks and the generic functions so they can gracefully handle it if there is not a remove search - but that should be done as a separate branch as a separate ticket.

Were you able to test this?  Just the check that is and not the action?

aschroed (approved):

PR #534, opened 2023-06-01
Dmichaels version of fix sh ct dec

dmichaels-harvard:
This has Andy's fix to prepare_static_headers_Chromatin_Tracing in checkcs/header_checks.py in the branch. And also some minor UI fixes related to inconsistencies with checks and actions definitions.

### DIFF
--- a/chalicelib_fourfront/checks/header_checks.py
+++ b/chalicelib_fourfront/checks/header_checks.py
@@ -314,7 +314,7 @@ def patch_static_headers_DNase_HiC(connection, **kwargs):
     return action
 
 #Chromatin Tracing
-@check_function()
+@check_function(action="patch_static_headers_Chromatin_Tracing")
 def prepare_static_headers_Chromatin_Tracing(connection, **kwargs):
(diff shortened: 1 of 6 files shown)


chalicelib_fourfront/checks/header_checks.py
```diff
@@ -314,11 +314,11 @@ def patch_static_headers_DNase_HiC(connection, **kwargs):
     return action
 
 #Chromatin Tracing
-@check_function()
+@check_function(action="patch_static_headers_Chromatin_Tracing")
```
netsettler, line 317: Why is it `patch` in the declaration and `prepare` in the definition name?

netsettler (approved):
A minor detail to check on, but nothing obviously blocking.

chalicelib_fourfront/checks/header_checks.py
dmichaels-harvard: The patch in `patch_static_headers_Chromatin_Tracing` is the name of the action function below (line 330).

PR #543, opened 2023-08-11
rewrite sync_users_oh to remove pandas and numpy

clarabakker:
Modified `sync_users_oh_check` and `sync_users_oh_status` to remove dependency on pandas (and also numpy). Updated lock file using `poetry update pandas`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ foursight
 Change Log
 ----------
 
+3.6.4
+=====
+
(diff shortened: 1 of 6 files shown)


clarabakker:
Yes, tested with a small sample sheet to check the formatting and then with the actual sheet with the libraries removed/lock file updated and installed again. I ran both the check and action.

aschroed (approved):

dmichaels-harvard (approved):
Thanks!

PR #548, opened 2023-09-21
Upgrade Python 3.11

dmichaels-harvard:


pyproject.toml
```diff
@@ -9,21 +9,24 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.7.1,<3.10"
-dcicutils = "^7.8.0"
+python = ">=3.7.1,<3.12"
```
netsettler, line 12: drop 3.7

netsettler (approved):
needs to drop 3.7 and workflow main-ci.yml might need a matrix tester. see other discussion on slack about whether we can afford such matrix testing.

pyproject.toml
dmichaels-harvard: Done

PR #551, opened 2023-10-30
Upgrade foursight to run repliseq v16.1

clarabakker:


aschroed (approved):
I might suggest bumping the version to 14.1.0 as this is updating a feature rather than fixing a bug but I leave that up to you.

aschroed (changes requested):
Looking again should not v16.1 be added to TSA-seq and NAD-seq  in wfr_utils?  Or am I not understanding how this works?

chalicelib_fourfront/checks/helpers/wfr_utils.py
```diff
@@ -189,11 +189,13 @@
...
+                          'RepliSeq_Pipeline_v16_step1',
+                          'RepliSeq_Pipeline_v16.1_step1'],
     # OFFICIAL - 1 STEP
     'Multi-stage Repli-seq': ['RepliSeq_Pipeline_v13.1_step1',
                               'RepliSeq_Pipeline_v14_step1',
-                              'RepliSeq_Pipeline_v16_step1'],
+                              'RepliSeq_Pipeline_v16_step1',
+                              'RepliSeq_Pipeline_v16.1_step1'],
     # Preliminary - Released to network
     'NAD-seq':       ['RepliSeq_Pipeline_v13.1_step1', 'RepliSeq_Pipeline_v14_step1', 'RepliSeq_Pipeline_v16_step1'],
```
aschroed, line 200: Should v16.1 be added here and in TSA-seq section above?

aschroed (approved):

chalicelib_fourfront/checks/helpers/wfr_utils.py
clarabakker: Good point!

PR #553, opened 2023-12-12
md5_status check update

aschroed:
This check has a throttle that limits the number of runs of these workflows that can be kicked off by the action to avoid going beyond the docker-hub pull limit of 200/6h.

Previously the way this was written was that the list of files checked was obtained by grabbing first n items from the search response where n is determined as (pull limit - the number of workflow_runs) created in the past 6 hours.  However, we just ran into the case where 1 submitter uploaded 100+ files (and the md5 was kicked for some but not all of them - due to the pull limit throttle ) and in the meantime before n was high enough to finish all the runs a different submitter uploaded metadata for ~1000 files but did not upload the actual files.  This meant that the same n files from the beginning of the list (those most recently submitted) got processed by the check and since they are still not uploaded the check is futilely spinning until they are.

This PR allows all the file items in the search_metadata response to be checked and how the appropriate number to run to avoid hitting the limit is determined (within the iterated results rather than limiting the number of items that are input to the iteration.

NOTE: that there is an (probably related) issue in that the check does not look to be triggered at all even though it is scheduled.  We thought that updating the docker that was pointed to in the Workflow Item had fixed this but the check still doesn't seem to be running as far as Will can tell.  However, the check appeared to run fine locally both before and after this update so I have no way to know if this change will resolve that.  I did also switch search_metadata to return as a generator if maybe there is a memory issue.

### DIFF
--- a/chalicelib_fourfront/checks/wfr_checks.py
+++ b/chalicelib_fourfront/checks/wfr_checks.py
@@ -214,7 +214,7 @@ def md5run_status(connection, **kwargs):
         check.status = 'WARN'
     if not_switched_status:
         check.allow_action = True
-        check.summary += 'Some files have wrong status and will re-run to update\n'
+        summary += 'Some files have wrong status and will re-run to update\n'
         msg = str(len(not_switched_status)) + ' file(s) have wrong status with a successful run to start'
         check.brief_output.append(msg)
         check.full_output['files_with_run_and_wrong_status_to_start'] = not_switched_status
--- a/chalicelib_fourfront/checks/wfr_checks.py
+++ b/chalicelib_fourfront/checks/wfr_checks.py
@@ -139,19 +139,24 @@ def md5run_status(connection, **kwargs):
     # if there are files, make sure they are not on s3
     no_s3_file = []
     running = []
-    missing_md5 = []
+    missing_md5_to_start = []
+    missing_md5_to_wait = []
     not_switched_status = []
+    not_switched_status_to_wait = []
     # multiple failed runs
     problems = []
     my_s3_util = s3Utils(env=connection.ff_env)
     raw_bucket = my_s3_util.raw_file_bucket
     out_bucket = my_s3_util.outfile_bucket
-    for a_file in res[:n_runs_available]:
+    for a_file in res:
         # lambda has a time limit (300sec), kill before it is reached so we get some results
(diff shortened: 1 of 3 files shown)


willronchetti (approved):
Seems like some of these changes are bug fixes ie: appending to summary that should be ported elsewhere?

chalicelib_fourfront/checks/wfr_checks.py
```diff
@@ -169,46 +174,66 @@ def md5run_status(connection, **kwargs):
...
+        msg = str(len(missing_md5_to_start)) + ' file(s) to start md5 run'
         check.brief_output.append(msg)
-        check.full_output['files_without_md5run'] = missing_md5
+        check.full_output['files_without_md5run_to_start'] = missing_md5_to_start
         check.status = 'WARN'
     if not_switched_status:
         check.allow_action = True
-        check.summary += ' Some files are have wrong status with a successful run'
-        msg = str(len(not_switched_status)) + ' file(s) are have wrong status with a successful run'
+        check.summary += 'Some files have wrong status and will re-run to update\n'
```
clarabakker, line 217: Should this be just `summary`? Apologies if I'm missing something

aschroed: Good catch!

chalicelib_fourfront/checks/wfr_checks.py
```diff
@@ -132,26 +132,31 @@ def md5run_status(connection, **kwargs):
     if lab:
         query += '&lab.display_title=' + lab
     # The search
-    res = ff_utils.search_metadata(query, key=my_auth)
+    res = ff_utils.search_metadata(query, key=my_auth, is_generator=True)
     if not res:
```
clarabakker, line 136: Also maybe naive, but does the generator returned in the previous line evaluate differently here? I am not sure what the generator will be if there are no results.

aschroed: Another great catch!  Thanks!

clarabakker (approved):
~~I tested the branch with a null query to see whether the conditional evaluated as expected with and without the generator, and I believe it does behave differently. Ultimately the check still correctly evaluates, but the generator does break that piece. ~~

Never mind, you were too speedy. I think this looks good!

PR #560, opened 2024-01-31
Bug fix for failing biorxiv update check

aschroed:
An exception was being thrown as the doi ID used for a pub contained a v# when submitted.  That doi was resolvable and the pub page looked fine.  However, when the biorxiv api was queried with that doi+v extension it returned an unexpected and handled response.  

Now this version bit is parsed out of the ID so api call works as expected
Added conditionals to be sure.
Also now tracking potentially misformatted doi values so they are reported and can be fixed if necessary.

### DIFF
--- a/chalicelib_fourfront/checks/wrangler_checks.py
+++ b/chalicelib_fourfront/checks/wrangler_checks.py
@@ -533,7 +533,7 @@ def biorxiv_version_update(connection, **kwargs):
         pubid = publication.get('ID')
         if not pubid.startswith('doi:'):
             continue
-        doi, misformatted = _remove_prefix_and_version_suffix_from_doi(publication['ID'])
+        doi, misformatted = _remove_prefix_and_version_suffix_from_doi(pubid)
         if misformatted:
             problem_ids.append(pubid)
         if not doi:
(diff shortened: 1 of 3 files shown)


chalicelib_fourfront/checks/wrangler_checks.py
```diff
@@ -512,12 +527,18 @@ def biorxiv_version_update(connection, **kwargs):
...
 
     items_to_update = []
+    problem_ids = []
     biorxiv_api = 'https://api.biorxiv.org/details/biorxiv/'
     for publication in current_biorxivs:
-        if not publication['ID'].startswith('doi:'):
+        pubid = publication.get('ID')
+        if not pubid.startswith('doi:'):
+            continue
+        doi, misformatted = _remove_prefix_and_version_suffix_from_doi(publication['ID'])
```
clarabakker, line 536: Pretty minor, but since you have `pubid`, can replace `publication['ID']` (i.e. `_remove_prefix_and_version_suffix_from_doi(pubid)`)

chalicelib_fourfront/checks/wrangler_checks.py
```diff
@@ -540,6 +561,15 @@ def biorxiv_version_update(connection, **kwargs):
     else:
         check.status = 'PASS'
         check.summary = check.description = 'All current bioRxiv Publications are up to date'
+
+    if problem_ids:
+        check.status = 'WARN'
+        check.summary = check.summary + f"\nThere are {len(problem_ids)} misformatted or problematic doi pub IDs"
```
clarabakker, line 567: For some reason, the newline shows up as just `\n` in the foursight check output (I noticed it in the doppelgänger check output). Okay to omit?

clarabakker (approved):
I think it looks good. I made two boring notes 🙂

chalicelib_fourfront/checks/wrangler_checks.py
aschroed: I don't think it will be added as a new line so might look funky but it already does anyway so removed \n

PR #561, opened 2024-02-08
Fix wfr_checks.md5run_status for calling any() on generator.

dmichaels-harvard:
Fix wfr_checks.md5run_status for bug where it was missing the first item in the result set because it was calling any() on a generator before iterating through it, which is destructive of the generator, i.e. causing to to move one item forwared.

### DIFF
--- a/chalicelib_fourfront/checks/wfr_checks.py
+++ b/chalicelib_fourfront/checks/wfr_checks.py
@@ -135,9 +135,11 @@ def md5run_status(connection, **kwargs):
     query += '&limit=' + limit
     # The search
     res = ff_utils.search_metadata(query, key=my_auth, is_generator=True)
-    if not any(res):
-        check.summary = 'All Good!'
-        return check
+    # 2024-02-08: Note that any() is destructive of the (res) generator;
+    # so this code was causing us to skip past the first result in the generator.
+    # if not any(res):
+    #    check.summary = 'All Good!'
+    #    return check
     # if there are files, make sure they are not on s3
     no_s3_file = []
     running = []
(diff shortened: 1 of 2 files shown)


aschroed (approved):

chalicelib_fourfront/checks/wfr_checks.py
```diff
@@ -147,10 +149,26 @@ def md5run_status(connection, **kwargs):
...
+    # this S3 access only if we have at least one result in the loop below.
+    # my_s3_util = s3Utils(env=connection.ff_env)
+    # raw_bucket = my_s3_util.raw_file_bucket
+    # out_bucket = my_s3_util.outfile_bucket
+    my_s3_util = None
+    while True:
+        try:
+            # TODO: Be nicer to have a generator wrapper to allow easily
+            # checking for any results without skipping past the first item.
+            a_file = next(res)
```
dmichaels-harvard, line 162: I don’t really like the way I did this, but best i could do without being more disruptive of the code for short term - though let me know what you think, if you have a better idea.

chalicelib_fourfront/checks/wfr_checks.py
```diff
@@ -147,10 +149,26 @@ def md5run_status(connection, **kwargs):
...
+            if not my_s3_util:
+                my_s3_util = s3Utils(env=connection.ff_env)
+                raw_bucket = my_s3_util.raw_file_bucket
+                out_bucket = my_s3_util.outfile_bucket
```
willronchetti, line 166: This you definitely can put outside the loop no?

chalicelib_fourfront/checks/wfr_checks.py
willronchetti: You can do this by making a second request with limit=0 and checking the "total" field in the (non-generator) response, and if there are results, then request the generator, otherwise exit

willronchetti (approved):
Fine to keep as is but I do have a suggestion for doing this more cleanly perhaps

chalicelib_fourfront/checks/wfr_checks.py
aschroed: I had initially thought to just get rid of the generator all together but we have a submitter just about to upload more than 1000 files that will need runs so best to either keep it the way you have it or use Will's approach.

chalicelib_fourfront/checks/wfr_checks.py
aschroed: I think no reason why not but if nothing is returned by the search then it's done for nothing - at least I guess that is what David was trying to avoid by checking and then creating if not there yet within the loop?

PR #564, opened 2024-03-14
Improved handling of user query for higlass items

clarabakker:
Add helper for three functions used by foursight checks querying for higlass items: `find_expsets_processed`, `files_requiring_higlass_item`, and
`find_expsets_otherprocessedfiles_requiring_higlass_items`. Queries formatted as strings will be corrected to lists (using input commas to split strings as necessary). A warning will be added to `check.brief_output` in this case.

### DIFF
--- a/chalicelib_fourfront/checks/higlass_checks.py
+++ b/chalicelib_fourfront/checks/higlass_checks.py
@@ -361,6 +361,9 @@ def find_files_requiring_higlass_items(connection, check_name, action_name, sear
         check.allow_action = False
         return check
 
+    elif search_queries.__class__ != list:
+        search_queries = [search_queries]
+
     # Add the fields we want to return.
     fields_to_include = '&field=' + '&field='.join((
         'accession',
--- a/chalicelib_fourfront/checks/higlass_checks.py
+++ b/chalicelib_fourfront/checks/higlass_checks.py
@@ -834,6 +834,9 @@ def find_expsets_processedfiles_requiring_higlass_items(connection, check_name, 
         check.allow_action = False
         return check
 
+    elif search_queries.__class__ != list:
+        search_queries = [search_queries]
+
     expsets_by_accession = {}
     # Use all of the search queries to make a list of the ExpSets we will work on.
     for query in search_queries:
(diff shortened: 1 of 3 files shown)


clarabakker:
I tested with properly formatted lists (1-2 items), multi-query strings with commas, single query strings, and empty lists.

chalicelib_fourfront/checks/higlass_checks.py
```diff
@@ -354,12 +354,7 @@ def find_files_requiring_higlass_items(connection, check_name, action_name, sear
     check.queries = []
     check.action = action_name
 
-    # If no search query was provided, fail
-    if not search_queries:
-        check.summary = check.description = "No search query provided, nothing to update."
-        check.status = 'PASS'
-        check.allow_action = False
-        return check
+    check, search_queries = verify_queries(check, search_queries, False)
```
aschroed, line 357: I don't believe you need to return the check from this function - you are passing a reference to an object which can be modified in the function and that change will be reflected without needing to return it.

aschroed (changes requested):
These functions are pretty messy but without overhauling this whole set of checks and as they mostly work fine not worth the effort to refactor.

I think it's really hard to know what is meant to be entered into the 'search_queries' parameter unless you know that the automated checks have the 'search_queries' stored in the full output and can be used as examples-but you need to know that first.

I was thinking that this initially would be easier to debug if we stored the actual full_query in the full_output (maybe without the field= ... part) but then if you tried to do what I suggested as a way to see what to pass into the parameter it would be even more confusing.  

Luckily, I think these get run rarely enough and by those of us who know where to look that again not worth making any changes there.

I do recommend adding another helper function like this:
`def get_search_results(connection, check, search_query):
    try:
        return ff_utils.search_metadata(search_query, key=connection.ff_keys)
    except Exception as e:
        check.full_output.setdefault('search_problems', []).append(e)
        return []`

And then I guess at least for the manual checks we might want to add a bit at the end of each to see if there is 'search_problems' included in the full output and if so set status to FAIL and disable the action?

aschroed (approved):

PR #565, opened 2024-03-25
Action can remove pending lab external-lab

clarabakker:
`wrangler_checks/users_with_pending_lab check` (and associated action):

- defaults to `WARN` (rather than using `PASS` to indicate `WARN`)
- removes external-lab from pending_lab as part of the action


clarabakker:
If the delete_field fails, is it okay to call the exception a "patch error"?

aschroed:
> If the delete_field fails, is it okay to call the exception a "patch error"?

Sure. It is in fact a patch operation from the apps perspective.

chalicelib_fourfront/checks/wrangler_checks.py
```diff
@@ -1105,7 +1105,7 @@ def users_with_pending_lab(connection, **kwargs):
     time.sleep(wait)
     check.action = 'finalize_user_pending_labs'
     check.full_output = []
-    check.status = 'PASS'
+    check.status = 'WARN'
```
aschroed, line 1108: I don't see anywhere that this check has a 'PASS' status.  Is that intentional?  Maybe in the ELSE along with `check.summary = 'No users found with pending_lab'`

aschroed (approved):

PR #567, opened 2024-04-05
Add organism when calling stepper

clarabakker:
Fix for usage of the stepper helper function: ATAC-seq metadata for processed files no longer defaults to human


aschroed (approved):
Looks good.

PR #569, opened 2024-04-23
Fix calls to get_es_metadata in checks/audit_checks.py to work when E…

dmichaels-harvard:
Fix calls to get_es_metadata in checks/audit_checks.py to work when ES_HOST_LOCAL is set.


aschroed (commented):
There are a few other cases when get_es_metadata is called - one in wrangler_checks and I think a couple more in es_checks?

chalicelib_fourfront/checks/helpers/es_utils.py
```diff
@@ -0,0 +1,14 @@
+import os
+from typing import Optional
+from dcicutils.es_utils import create_es_client
+from dcicutils import ff_utils
+
+def get_es_metadata(*args, **kwargs):
+    if (kwargs.get("es_client", None) is None) and ((es_host_local := _get_es_host_local()) is not None):
+        es_client = create_es_client(es_host_local, use_aws_auth=True)
+        return ff_utils.get_es_metadata(*args, **kwargs, es_client=es_client)
+    return ff_utils.get_es_metadata(*args, **kwargs)
```
willronchetti, line 10: This should go in dcicutils for sure

willronchetti (approved):
Should double check for other references to repair, and also refactor into utils

PR #571, opened 2024-05-13
Hi-C update to 0.3.0

clarabakker:
Add new workflow versions for Hi-C and new version tags for each Hi-C experiment type.


aschroed (approved):
This looks good.  

Perhaps a minor rather than patch version bump to 4.7.0?

PR #577, opened 2024-08-05
Get info about workflows and pipelines from db instead of hard coding

aschroed:
Needs careful review and testing
* Refactor of wfr_utils to utilize properties and tags on updated schemas for workflow and experiment_type to generate info about accepted versions of workflows and pipelines
* Updated wfr_checks and wfr_encoded_checks to utilize the new info from db
* provided additional keyword args to override the accepted versions (useful for testing or some cases where re-runs of updated pipelines are needed)
* Fix bug that allowed more than 2 errored fastq_first_line runs happen

### DIFF
--- a/chalicelib_fourfront/checks/helpers/wfr_utils.py
+++ b/chalicelib_fourfront/checks/helpers/wfr_utils.py
@@ -156,7 +156,6 @@ def check_indexing(check, connection):
     # wait for random time
     wait = round(random.uniform(0.1, random_wait), 1)
     time.sleep(wait)
-    # TEMPORARILY DISABLE ALL PIPELINE RUNS
     env = connection.ff_env
     if env in [FF_PRODUCTION_IDENTIFIER, FF_STAGING_IDENTIFIER]:
         health = ff_utils.get_health_page(ff_env=env)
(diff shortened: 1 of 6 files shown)


chalicelib_fourfront/checks/helpers/wfr_utils.py
```diff
@@ -540,9 +352,9 @@ def get_wfr_out(emb_file, wfr_name, key=None, all_wfrs=None, versions=None,
 
     error_at_failed_runs = error_threshold
     # you should provide key or all_wfrs
-    assert key or all_wfrs
-    if wfr_name not in workflow_details:
-        assert wfr_name in workflow_details
+    assert key
+    workflow_details = get_workflow_details(key)
+    assert wfr_name in workflow_details
```
clarabakker, line 357: Ran into an error at this point in a check running--it looks like `workflow_details` is just `{}`. I will keep exploring!

aschroed: This makes sense as the properties that were added to the Workflow and ExperimentType items are not in webdev and I guess the version of fourfront there is likely prior to the schema changes so they can't be added.  Note that for testing particular workflows you can pass in this info as a parameter but that kind of defeats some of the purpose of testing.  I will consider options for making this backward compatible so it can be used on webdev without errors.

clarabakker: Oh, of course!

clarabakker:
Okay, I managed to get a QC workflow to run and patch on webdev, as well as a processing workflow using the stepper function. It was a little tricky to get working on webdev for the processing pipeline, but it definitely works.

I guess I'll turn the review back to you for the addition to prevent fastq-first-line from running over 2 errors and to bump versions.

PR #579, opened 2024-08-13
Update entrez_gene_id validation check

aschroed:
To work with batch eutils fetch.

Note: this was tested locally using a beta version of foursight-core that allows short term cred use in local-check-execution but this PR uses existing foursight-core specification in pyproject.toml.

### DIFF
--- a/chalicelib_fourfront/checks/wrangler_checks.py
+++ b/chalicelib_fourfront/checks/wrangler_checks.py
@@ -1053,38 +1053,47 @@ def validate_entrez_geneids(connection, **kwargs):
     problems = {}
     timeouts = 0
     search_query = 'search/?type=Gene&limit=all&field=geneid'
+    import pdb; pdb.set_trace()
+
     genes = ff_utils.search_metadata(search_query, key=connection.ff_keys)
     if not genes:
         check.status = "FAIL"
         check.description = "Could not retrieve gene records from fourfront"
         return check
     geneids = [g.get('geneid') for g in genes]
-
-    query = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=gene&id={id}"
-    for gid in geneids:
+    ID_BATCH_SIZE = 150
+    query = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=gene&id={id}&rettype=uilist&retmode=text"
+    for i in range(0, len(geneids), ID_BATCH_SIZE):
+        batch = geneids[i: i + ID_BATCH_SIZE]
+        gids = ', '.join(batch)
         if timeouts > 5:
             check.status = "FAIL"
             check.description = "Too many ncbi timeouts. Maybe they're down."
             return check
-        gquery = query.format(id=gid)
+        gquery = query.format(id=gids)
         # make 3 attempts to query gene at ncbi
         for count in range(3):
             resp = requests.get(gquery)
             if resp.status_code == 200:
+                try:
+                    rtxt = resp.text
+                    if rtxt.startswith('Error'):
+                        problems['Error response'].extend(batch)
+                    retids = [i for i in resp.text.split('\n') if i]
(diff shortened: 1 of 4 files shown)


chalicelib_fourfront/checks/wrangler_checks.py
```diff
@@ -1059,32 +1059,39 @@ def validate_entrez_geneids(connection, **kwargs):
         check.description = "Could not retrieve gene records from fourfront"
         return check
     geneids = [g.get('geneid') for g in genes]
-
-    query = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=gene&id={id}"
-    for gid in geneids:
+    ID_BATCH_SIZE = 150
+    query = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=gene&id={id}&rettype=uilist&retmode=text"
```
clarabakker, line 1063: I am not very familiar with this check, so I was testing the queries to see the response. I may be misunderstanding or not testing correctly, but it seems like this query just gives me back whatever numbers I put in my test ID, with no other info, regardless of whether it produces an error for the original query. Do you also see this behavior?

aschroed: That is the intended behavior I believe.  The ids are passed in as a comma separated list. Have you tried to include non-valid geneids?  Invalid IDs should not be included in the result list and then that list is checked against the input list and anything in input list not in the output is reported as problematic.  At least that is the idea.

clarabakker: Yes, I mean that any string of digits will appear in the list, even invalid.

aschroed: Ah right you are!  Good catch.  If you put in something that is not a number you get something useful but it returns even invalid geneids as numbers.  OK I will need to fix that.

aschroed:
@clarabakker this has now been updated - I reverted back to fetching genes iteratively but moved the sleep so while the check now takes longer it should not result in 429 errors from ncbi that I believe are the cause of the issue.

Additionally - before the check just made sure some record was returned by ncbi gene database but now it tries to check and make sure the gene status is 'alive'.

However, not all gene records have a status - don't ask me why - so I added an ignore list that is stored in the result and the ability to modify ids in that list - either adding or removing.

Hopefully this should now cover the bases and actually do some validation without erroring.  If gene number gets too high for the lambda limit we may need to revisit but hopefully this will keep us in good shape for a while. 

Please have a look and see what you think.

chalicelib_fourfront/checks/wrangler_checks.py
```diff
@@ -1051,14 +1051,53 @@ def find_entrez_gene_status(report):
...
+    last_result = check.get_primary_result()
+    # hack to add empty full_output for first time it is used
+    # last_result['full_output'] = {}
+    # if last one was fail, find an earlier check with non-FAIL status
+    it = 0
+    while last_result['status'] == 'ERROR' or not last_result['kwargs'].get('primary'):
+        it += 1
+        # this is a daily check, so look for checks with 12h iteration
+        hours = it * 12
+        last_result = check.get_closest_result(diff_hours=hours)
```
clarabakker, line 1073: I am getting this far in my test run when I run into an exception (before reaching the 100 hours) in `get_closest_result`--maybe it needs to be caught so it can continue? The exception: `Could not find closest non-ERROR result for prefix: validate_entrez_geneids. Attempted with 12 diff hours and 0 diff mins.`

It also gave me a couple other questions: do we have a successful result within 100 hours? Or how do we resolve that 100 hour error if it occurs?

clarabakker (approved):
Looks good to me! I didn't encounter the invalid literal error, so I can't speak to that, but the sorting and other finicky bits I mentioned all seem resolved now.

RahiNav (approved):
Looks good. No pending issues during testing.

PR #580, opened 2024-10-02
Ajs fix bugs

aschroed:
Fixed bugs in 3 different checks:
1. in check that looks for reciprocal and complete relationships for the 'grouped with' relationship type the formation of groups was somewhat relationship type agnostic so there were paired fastq files that should not get the grouped with relationship added to groups if other files linked to the pair had that relationship - eg. index reads or barcodes.  Added a filter to avoid this incorrect grouping
2. entrez_gene_id validation would throw and exception if there was no response at all from the ncbi request so added a check to make sure that if that was the case it would be dealt with gracefully
3. SPIN state files get a tag and higlass_defaults for color a multivec added - the check that dealt with finding files that needed this info updated was very wonky so this PR refactors that code.

### DIFF
--- a/chalicelib_fourfront/checks/wrangler_checks.py
+++ b/chalicelib_fourfront/checks/wrangler_checks.py
@@ -1151,12 +1151,13 @@ def validate_entrez_geneids(connection, **kwargs):
         gquery = query.format(id=gid)
         # make 3 attempts to query gene at ncbi
         for count in range(3):
+            resp = None
             try:
                 resp = requests.get(gquery)
             except Exception:
                 pass  # after 3 times will hit conditional below
             time.sleep(0.334)
-            if resp.status_code == 200:
+            if resp and resp.status_code == 200:
                 break
             if count == 2:  # third try without 200
                 timeouts += 1
--- a/chalicelib_fourfront/checks/wrangler_checks.py
+++ b/chalicelib_fourfront/checks/wrangler_checks.py
@@ -1712,7 +1712,6 @@ def states_files_without_higlass_defaults(connection, **kwargs):
         acc = a_res.get('accession')
         has_higlass_defaults = False
         has_tag = False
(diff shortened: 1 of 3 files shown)


RahiNav:
Hi, the local tests for grouped_with_file_relation_consistency and  validate_entrez_geneids came back as expected. 
For the grouped_with check: {
    "name": "grouped_with_file_relation_consistency",
    "summary": "All 'grouped with' file relationships are consistent",
    "description": "All 'grouped with' file relationships are consistent",
    "status": "PASS",......}
For gene_id check:{
    "name": "validate_entrez_geneids",
    "summary": "",
    "description": "GENE IDs are all valid",
    "status": "PASS",.....}

RahiNav (approved):

chalicelib_fourfront/checks/wrangler_checks.py
```diff
@@ -1702,24 +1703,41 @@ def states_files_without_higlass_defaults(connection, **kwargs):
...
-                check.full_output['problematic_files'][a_res['accession']] = 'missing state tag'
-            else:
-                check.full_output['to_add'][a_res['accession']] = a_res["tags"]
-
-    if check.full_output['to_add']:
+        uid = a_res.get('uuid')
+        acc = a_res.get('accession')
+        has_higlass_defaults = False
+        has_tag = False
+        import pdb; pdb.set_trace()
```
utku-ozturk, line 1715: @aschroed First of all, I apologize for commenting on this PR even though I wasn't assigned to it, but this line caught my attention. Should it stay like this, or was it overlooked? That's why I wanted to mention it.

aschroed: Thanks @utku-ozturk - overlooked and you should always feel free to comment - especially if you find something like this.

PR #583, opened 2024-10-08
Update opf_status_mismatch check

aschroed:
- opf_status_mismatch check will now ignore mismatches in items that have a tag added 
- added &tags!=skip_processing to file search in bed2beddb check

The opf_status check was tested locally.
Did not test the bed2beddb check.

### DIFF
--- a/chalicelib_fourfront/checks/audit_checks.py
+++ b/chalicelib_fourfront/checks/audit_checks.py
@@ -734,6 +734,9 @@ def check_opf_status_mismatch(connection, **kwargs):
     '''
     check = CheckResult(connection, 'check_opf_status_mismatch')
 
+    # list of uuids to filter out as they have a tag to ignore them
+    tagged2ignore = get_items_with_ignore_tags(connection.ff_keys)
+
     opf_set = ('search/?type=ExperimentSet&other_processed_files.title%21=No+value&field=status'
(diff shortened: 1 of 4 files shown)


chalicelib_fourfront/checks/wfr_checks.py
clarabakker, line 1: I think you may also need to change the query at line 661 (unless you only want to check after the run failed)

aschroed: Yep good catch - thanks!

RahiNav (approved):

chalicelib_fourfront/checks/audit_checks.py
```diff
@@ -782,7 +797,7 @@ def check_opf_status_mismatch(connection, **kwargs):
             file_list.extend([item for exp in result.get('experiments_in_set', [])
                               for fileset in exp['other_processed_files']
                               for item in fileset['files'] if fileset['title'] == title])
-            statuses = set([opf_status_dict[f['uuid']] for f in file_list])
+            statuses = set([opf_status_dict[f['uuid']] for f in file_list if f.get('uuid') not in tagged2ignore])
             # import pdb; pdb.set_trace()
```
clarabakker, line 801: Not from you, but could be removed if you'd like

clarabakker (approved):

PR #585, opened 2024-11-12
Remove rate limit for md5_status check

clarabakker:
Remove rate limiting (and related variables) for md5_status check.

No md5-related checks should use rate limiting after this change.


aschroed (approved):
Looks good!

PR #587, opened 2025-01-07
Remove restricted status from 2 checks

aschroed:
In order to prevent triggering fastq_firstline or bamQC runs on 'restricted' files that have not been uploaded and therefore will fail the runs the searches in those checks have been modified to remove files with 'restricted' status from the results.


clarabakker (approved):
Looks good to me!

RahiNav (approved):

PR #588, opened 2025-01-28
Deduplicate control TAs in ChIP-seq input JSON

clarabakker:
For ChIP-seq sets with 2+ identical control tagAlign files, replace multi-file control list with a one-item list.

For example:
`s2_input_files['chip.ctl_tas'] = ['fileA', 'fileA', 'fileA']` will be replaced with `s2_input_files['chip.ctl_tas'] = ['fileA']`

Files are also renamed accordingly. No changes made in the case that some but not all control files are identical.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ Change Log
 4.9.10
 =====
 
-`PR`_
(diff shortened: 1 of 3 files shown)


aschroed (approved):
Looks good.

PR #590, opened 2025-03-18
Audit check 'bug fixes'

aschroed:
* Fix for the other processed files mismatch check was reporting a mismatched higlass viewconf even though that item had the ignore_status_mismatch tag.  
* Fix for 2 unlinked files (i.e. not linked to WFRs as outputs as they were part of supplementary files for 0.3 Hi-C pipe run) that were being flagged in check for restricted processed files that were released - added a tag for the files and included a check for said tag in foursight

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ foursight
 Change Log
 ----------
 
+
+4.9.15
+======
+
+* fix for opf status mismatch audit to deal correctly with highglass viewconf mismatch even if ignore tag was present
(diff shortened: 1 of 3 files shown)


RahiNav (approved):

clarabakker (approved):
Works for me. I noticed a line (755) commented out which seems like it was meant to check the higlass views. (It looks like it has always been a comment, though: https://github.com/4dn-dcic/foursight/commit/78f57e3b8a058776cbcf43239ba2fe4b5d223db5#diff-a1674260d181e0548d79b9c4afc1b26d67a9d4f6737e08b0cf5b449015e2b2f4R755)

It could be removed, but not hurting anything.

PR #594, opened 2025-04-24
Ajs 250423 chk ss url

aschroed:
Updated header with new user agent info for help page url check


utku-ozturk (approved):
looks good to me!

PR #605, opened 2025-07-31
Ajs 250731 fix not uploaded bug

aschroed:
Fix for a bug that was introduced in the check for uploaded files that lacked an md5 runs.  
2 changes:
- deal properly with processed files
- rename the keys for storing the problematic files as they were mismatched


RahiNav:
@aschroed Makes sense and looks good!

RahiNav (approved):

